PR #7, opened 2019-07-08
Testing Suite Setup

jgreben:
Adds Rspec, rubocop-rspec, capybara (with empty features dir), Webdriver, Webmock, and sets up rubocop rake task


spec/spec_helper.rb
```diff
@@ -0,0 +1,100 @@
+# frozen_string_literal: true
+
+require 'webmock/rspec'
+
+WebMock.disable_net_connect!(allow_localhost: true)
```
camillevilla:
```suggestion
WebMock.disable_net_connect!(allow_localhost: true, allow: 'chromedriver.storage.googleapis.com')
```
We'll want to add this exception so that `webdrivers` can request chromedriver.

jgreben: 👌

PR #24, opened 2019-07-08
Google analytics

### LINKED ISSUE #2: Hook up Google Analytics
Add basic Google Analytics integration to the application, using account `UA-7219229-36`.

Also be sure to enable [IP address anonymization configuration](https://support.google.com/analytics/answer/2763052?hl=en).

shelleydoljack:
Closes #2


config/settings.yml
```diff
@@ -0,0 +1 @@
+GOOGLE_ANALYTICS_ID:
```
cbeer:
```suggestion
GOOGLE_ANALYTICS_ID: UA-7219229-36
```

I think it's probably fine to put this into the global settings.

mejackreed: By default that would add analytics on all our environments (dev, uat) to the same spot (i think)

Issue #42, opened 2019-07-09
Add recaptcha for feedback form for non-logged in users


mejackreed:
Split out of #20

PR #43, opened 2019-07-09
Adds feedback form and flash partial

### LINKED ISSUE #20: Add a Feedback widget
Add yet another SUL feedback widget. Maybe give some thought towards creating a pattern that's useful for other applications, but creating the one feedback form to rule them all is not the priority.

Crib from Searchworks, including the CAPTCHA for unauthenticated users.

This may also be useful as a way to gather feedback during the development cycle.

### LINKED ISSUE #38: Add a consistent place to display flash messages
Many user actions will trigger flash messages:

Screen Shot 2019-07-09 at 08 25 12
Screen Shot 2019-07-09 at 08 25 19
Screen Shot 2019-07-09 at 08 25 26
Screen Shot 2019-07-09 at 08 25 37

(These examples are all successes, but presumably failures or error reporting will be handled similarly) 

We should have a consistent way to display these to the user. This may be part of the layout, or a shared partial that can get reused.

mejackreed:
This PR fixes #20 and fixes #38 

I've spun out #42 until we have users / logged in capabilities

This ports over from SearchWorks (and EarthWorks BS4) feedback form infrastructure + tests. Part of that is the JavaScript based submission.

feedback


app/views/shared/feedback_forms/_form.html.erb
```diff
@@ -0,0 +1,48 @@
...
+          </div>
+        </div>
+        <div class="form-group row">
+          <%= label_tag(:to, 'Your email', class:"col-sm-3 col-form-label") %>
+          <div class="col-sm-9">
+            <%= email_field_tag :to, "", class:"form-control" %>
+          </div>
+        </div>
+        <div class="form-group row">
+          <div class="col-sm--3 col-sm-9">
```
camillevilla:
```suggestion
          <div class="col-sm-3 col-sm-9">
```
one lil' typo

mejackreed: Nice catch.. I've actually just pushed up removing `col-sm-3` all together.

PR #45, opened 2019-07-10
Add some basic scss configs

### LINKED ISSUE #11: Add SASS configuration for application-specific styles (headers, fonts, etc)
@jvine to provide styles for:

- [ ] page header links
- [ ] page nav buttons
- [ ] text headers
- [ ] body text
- [ ] table headers
- [ ] responsive text sizes
- [ ] modal text
- [ ] ALL THE THINGS

jkeck:
Closes #11 

Screen Shot 2019-07-10 at 10 08 08 AM


app/assets/stylesheets/bootstrap-variables.scss
```diff
@@ -0,0 +1,8 @@
+// Override bootstrap variables here
+
+// Headings
+$font-size-h1: 2.323em;
+$font-size-h2: 1.793em;
+$font-size-h3: 1.383em;
+$font-size-h4: 1.215em;
```
jkeck: These font sizes aren't actually working...

Screen Shot 2019-07-09 at 10 22 44 PM

We are explicitly setting various `h*` declarations, so maybe we should just set those there.  (the font-weight below does work, soooo)

app/assets/stylesheets/bootstrap-overrides.scss
```diff
@@ -0,0 +1,59 @@
...
+.btn-primary {
+  @include button-variant(
+    $sul-btn-primary-color,
+    darken($sul-btn-primary-color, 7.5%),
+    darken($sul-btn-primary-color, 10%),
+    lighten($sul-btn-primary-color, 5%),
+    lighten($sul-btn-primary-color, 10%),
+    darken($sul-btn-primary-color, 30%)
+  );
+}
```
jkeck: I'm surprised that we have to do this, but overriding the various `btn-*` variables doesn't appear to work in BS4. Happy to explore other alternatives if anybody is aware of them.

mejackreed: I think you have to either do this, or override the colors in the variables (if I remember correctly).

https://github.com/sul-dlss/earthworks/blob/master/app/assets/stylesheets/bootstrap-variables.scss#L4

PR #61, opened 2019-07-10
Adds checkouts tab

### LINKED ISSUE #18: Add Checkouts tab displaying checked out material
- [x] retrieve checkouts for a user using the Symphony API (Prerequisite: #13 )
- [x] translate internal codes (like library/location) to better user facing labels 
- [x] display them (Prerequisite: #16 (figuring out our generic data tables approach)) on the page
- [x] add a collapsable section with more information
- [x] provide a link to the item in Searchworks (#77 )
- [x] set an appropriate background color depending on the items status
- [x] pull recalled items to the top of the page
- [ ] make the data sortable using the sort widget (and, optionally, using the column headers)

## Mobile
Screen Shot 2019-07-08 at 15 48 01

## Desktop
Screen Shot 2019-07-08 at 15 47 50

shelleydoljack:
Part of #18

### DIFF
--- a/app/views/checkouts/index.html.erb
+++ b/app/views/checkouts/index.html.erb
@@ -3,3 +3,18 @@
 <h1>Checkouts</h1>
 
 <%= render 'shared/navigation' %>
+
+<% @checkouts.each do |checkout| %>
+  <div class="row">
+    <%= checkout.status %>
+    <%= checkout.overdue? %>
+    <%= checkout.due_date %>
+    <%= checkout.title %>
+    <%= checkout.author %>
+    <%= checkout.call_number %>
+    <%= checkout.checkout_date %>
+    <%= checkout.library %>
+    <%= checkout.recalled_date %>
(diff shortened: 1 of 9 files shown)


app/views/checkouts/index.html.erb
```diff
@@ -3,3 +3,17 @@
 <h1>Checkouts</h1>
 
 <%= render 'shared/navigation' %>
+
+<% @checkouts.each do |checkout| %>
+  <div class="row">
+    <%= checkout['fields']['status']%>
```
jkeck: It looks like this isn't accessible from the `Checkout` class.  It seems like there are accessor in that class for each of these pieces of data.  Perhaps you for forgot to push up a commit where you changes this view from accessing the json directly to model accessors?

cbeer: Thanks, it looks like I dropped a commit.

PR #93, opened 2019-07-12
Retrieve request/hold data from Symphony and poke it through the requests controller

### LINKED ISSUE #87: Retrieve user request data from Symphony and wire it up to the `RequestsController`
Part of #50

- [x] add an API call to `SymphonyClient` to get the requests for a given `patron_key`

The spreadsheet suggests it may be very similar to checkouts, but pulling from `holdRecordList{*}` instead of `circRecordList`.

- [x] Call the new method from the `RequestsController`
- [x] Wrap the list of requests in an application model (similar to `Checkout`) to make the Symphony data easier to access within our application

jkeck:
Closes #87

Screen Shot 2019-07-12 at 10 03 34 AM

### DIFF
--- a/app/models/request.rb
+++ b/app/models/request.rb
@@ -1,0 +1,60 @@
+# frozen_string_literal: true
+
+# Model for requests in Symphony
+class Request
+  attr_reader :record
+
+  def initialize(record)
+    @record = record
+  end
+
+  def key
+    record['key']
+  end
+
+  def status
+    fields['status']
+  end
+
+  def catkey
+    fields['item']['fields']['bib']['key']
+  end
+
+  def title
+    bib['title']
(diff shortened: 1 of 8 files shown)


app/models/request.rb
```diff
@@ -0,0 +1,60 @@
...
+    record['key']
+  end
+
+  def status
+    fields['status']
+  end
+
+  def catkey
+    fields['item']['fields']['bib']['key']
+  end
```
jkeck: Note, I haven't done an exhaustive look over the designs, in particular with the different flavors of dates that we want to use vs the name that's used in the API.  I figured this would be a good thing to pair up on as we start getting the data into the UI.

Issue #53, opened 2019-07-10
Add patron privileges information to the summary page


cbeer:
SUNETID accounts:
<img width="388" alt="Screen Shot 2019-07-10 at 08 06 03" src="https://user-images.githubusercontent.com/111218/60980572-9147eb00-a2e9-11e9-9faa-051f57af974f.png">

Paid accounts:
<img width="367" alt="Screen Shot 2019-07-10 at 08 06 19" src="https://user-images.githubusercontent.com/111218/60980594-9ad15300-a2e9-11e9-8af9-d95ecc0efcae.png">

Paid accounts also have a check out quota:
<img width="351" alt="Screen Shot 2019-07-10 at 08 06 54" src="https://user-images.githubusercontent.com/111218/60980631-b177aa00-a2e9-11e9-89b9-8c0f9b9e228c.png">

Translation table for patron status:

```
    'COLLECTION' => 'Blocked',
    'BARRED' => 'Blocked',
    'BLOCKED' => 'Blocked',
    'OK' => 'OK',
    'DELINQUENT' => 'OK'
```

jvine:
For BARRED, I'd like to try a translation of "Contact us" instead of "Blocked".

<img width="98" alt="Screen Shot 2019-07-17 at 11 44 18 AM" src="https://user-images.githubusercontent.com/6945691/61402581-46e5e180-a888-11e9-9fce-d88b54074846.png">

The user can clear the BLOCKED status by paying their fines. They can't do anything about BARRED _except_ contact Access Services. I'd like that to be visible in the nav, not just the summary page.

cbeer:
Remaining work split to #216

PR #100, opened 2019-07-12
Adds patron status and expiration logic

shelleydoljack:
Part of #53
Next steps are separating patron types (SUNetID, fee borrower, etc.).
Paired with Camille and Tony on this PR.

### DIFF
--- a/app/controllers/summaries_controller.rb
+++ b/app/controllers/summaries_controller.rb
@@ -6,7 +6,9 @@ class SummariesController < ApplicationController
 
   # GET /summaries
   # GET /summaries.json
-  def index; end
+  def index
+    @response = symphony_client.patron_info(current_user.patron_key)
+  end
 
   private
 
--- a/app/controllers/summaries_controller.rb
+++ b/app/controllers/summaries_controller.rb
@@ -7,7 +7,7 @@ class SummariesController < ApplicationController
   # GET /summaries
   # GET /summaries.json
   def index
-    @response = symphony_client.patron_info(current_user.patron_key)
+    @patron = symphony_client.patron_info(current_user.patron_key)
   end
 
   private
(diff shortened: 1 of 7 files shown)


app/controllers/summaries_controller.rb
```diff
@@ -6,7 +6,9 @@ class SummariesController < ApplicationController
 
   # GET /summaries
   # GET /summaries.json
-  def index; end
+  def index
+    @response = symphony_client.patron_info(current_user.patron_key)
```
cbeer: Should we call the instance variable `@patron` for clarity?

app/models/patron.rb
```diff
@@ -12,6 +20,27 @@ def key
...
+  def status
+    if expired?
+      'Expired'
+    else
+      PATRON_STANDING.fetch(fields['standing']['key'], '')
+    end
+  end
+
+  def expired?
+    expired_date = fields['privilegeExpiresDate']
```
cbeer: Maybe we could extract `expired_date` to its own method, and have it do the same sort of date parsing we do in e.g. `Checkout#recalled_date`?

app/models/patron.rb
```diff
@@ -12,6 +20,27 @@ def key
...
+    expired_date.present? ? Time.zone.now.strftime('%Y-%m-%d') >= expired_date : false
```
cbeer: If nothing else, it'd let us do a direct comparison here, instead of a string compare, maybe something like:

```suggestion
    expired_date && Time.zone.now >= expired_date
```

spec/models/patron_spec.rb
```diff
@@ -51,4 +69,33 @@
       expect(patron.email).to be_nil
     end
   end
+
+  describe '#expired?' do
```
cbeer: We should probably have additional tests for the other logic branches:

a) a privileges date in the future
b) no privileges date

PR #106, opened 2019-07-14
Contact access services

### LINKED ISSUE #36: Contact Access Services
Add a link to contact access services:
Screen Shot 2019-07-09 at 08 15 45

The telephone number can also be added at this time (presumably as a `tel:`-protocol link?)

The Contact Access Services form appears in a modal:

Screen Shot 2019-07-09 at 08 16 37

If we have an authenticated user, we can populate the `from` with the users information, otherwise it should have a similar behavior to the feedback form. ~Unauthenticated users should get the same CAPTCHA challenge as on the feedback form (#20).~

Note that a very similar contact form will also be used to contact libraries as well (see #37).

jgreben:
Fixes #36 

Adds Contact Access Services modal form.


jgreben:
Codeclimate is unhappy because we added a couple extra lines to a javascript method 👆

app/assets/javascripts/application.js
```diff
@@ -17,4 +17,5 @@
 //= require popper
 //= require list.min.js
 //= require bootstrap-sprockets
+//= require bootstrap/modal
```
mejackreed: Is this require needed here? I think `bootstrap-sprockets` gives us all of the things?

app/assets/stylesheets/modules/top_navbar.scss
```diff
@@ -38,8 +37,36 @@
...
 }
 
+.navbar {
+  max-height: 120px;
+
+  .header-bar {
+    flex-wrap: wrap !important;
+  }
+
+  #topnav-container {
```
mejackreed: Is this redundant styling (or unexecuted)? I'm seeing similar things on LN3 but without the qualifying parent classes

app/assets/stylesheets/modules/top_navbar.scss
```diff
@@ -38,8 +37,36 @@
  }
 }
 
+.navbar {
+  max-height: 120px;
+
+  .header-bar {
```
mejackreed: I'm not able to determine where this gets applied. Also should it live in `header_navbar.scss` if it is needed?

app/assets/stylesheets/modules/top_navbar.scss
```diff
@@ -38,8 +37,36 @@
...
+  }
+}
+
 @include media-breakpoint-up(xs) {
   #topnav {
     width: 100%;
   }
 }
+
+@media (min-width: 768px) {
```
mejackreed: Can we use the bootstrap defined breakpoints here? https://getbootstrap.com/docs/4.3/layout/grid/#example-usage

app/assets/stylesheets/modules/top_navbar.scss
```diff
@@ -38,8 +37,36 @@
...
+  .navbar {
+    max-height: 70px;
+  }
+
+  .navbar-nav {
```
mejackreed: Should these be scoped to `#topnav` or live in `header_navbar.scss` and scoped there? Or are they more globally applied?

config/settings/test.yml
```diff
@@ -4,3 +4,5 @@ symws:
   login_params:
     login: 'fake_user'
     password: 'some_password'
+
+ACCESS_SERVICES_EMAIL: greencirc@stanford.edu
```
mejackreed: Should this just be in `config/settings.yml` ?

app/views/shared/contact_forms/_form.html.erb
```diff
@@ -0,0 +1,24 @@
...
+</div>
+<div class="modal-body offset-sm-1">
+  <%= form_tag contact_path, method: :post, class: 'contact-form', role: 'form' do %>
+    <div class="row">
+      <%= hidden_field_tag :url, request.referer %>
+      <div class="col-sm-12">
+        From: <%= patron.display_name %>
+      </div>
+      <div class="col-sm-12">
+        <!--TODO: Get Access Services info from configuration? -->
```
mejackreed: Is this TODO taken care of now?

app/views/shared/contact_forms/_form.html.erb
```diff
@@ -0,0 +1,24 @@
+<div class="modal-header offset-sm-1">
+  <h3 id="contactForm">Contact Access Services</h3>
+</div>
+<div class="modal-body offset-sm-1">
+  <%= form_tag contact_path, method: :post, class: 'contact-form', role: 'form' do %>
+    <div class="row">
+      <%= hidden_field_tag :url, request.referer %>
```
mejackreed: This might not work when triggered from JavaScript. You could include this partial: https://github.com/sul-dlss/mylibrary/blob/master/app/views/shared/feedback_forms/_reporting_from.html.erb

or the appropriate markup to have the JavaScript do it for us.

app/assets/javascripts/application.js
jgreben: Yep looks like `bootstrap-sprockets` has what we need. I will remove the extra require.

app/assets/stylesheets/modules/top_navbar.scss
jgreben: Unexecuted. That id no longer exists. Removing.

app/assets/stylesheets/modules/top_navbar.scss
jgreben: Yep, it no longer appears to be needed. Removing.

app/assets/stylesheets/modules/top_navbar.scss
jgreben: It does not appear to be needed at all right now, so I am just removing it. I will dig into the using bootstrap breakpoints later if needed...

app/assets/stylesheets/modules/top_navbar.scss
jgreben: Also does not appear to be needed at all right now, so I am just removing it.

app/views/shared/contact_forms/_form.html.erb
jgreben: I just rendered the existing partial.

Issue #98, opened 2019-07-12
Patron info for Fee Borrowers


saseestone:
For Patron info: Symphony understands types of users using the "User Profile" in the user record.

User Profiles for Fee borrowers:
MXFEE
MXFEE-BUS
MXFEE-LAW
MXFEE-NO25

camillevilla:
Info consolidated in #118

PR #112, opened 2019-07-15
Adds fee borrower.

shelleydoljack:
Part of #53


camillevilla:
This PR adds the notion of a `patron_type`, which is mapped from Symphony `profile`s. For example, the notion of a "fee borrower" maps to several Symphony `profile`s (see #98). 

## SUNet user
sunet-user

## Fee borrower
fee-borrower-user-showing-type

PR #113, opened 2019-07-15
Get the Searchworks link to align with the rest of the expanded data

cbeer:
Screen Shot 2019-07-15 at 11 59 02

### DIFF
--- a/app/views/checkouts/_checkout.html.erb
+++ b/app/views/checkouts/_checkout.html.erb
@@ -39,8 +39,13 @@
       <dt class="col-5">Source:</dt>
       <dd class="col-5"><%= Mylibrary::Application.config.library_map[checkout.library] || checkout.library %></dd>
     </dl>
-    <%= link_to Settings.sw.url + checkout.catkey, target: '_blank' do %>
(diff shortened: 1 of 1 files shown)


app/views/checkouts/_checkout.html.erb
```diff
@@ -39,8 +39,13 @@
...
-    <%= link_to Settings.sw.url + checkout.catkey, target: '_blank' do %>
-      <%= sul_icon :'sharp-open_in_new-24px' %> View in SearchWorks
-    <% end %>
+    <div class="row justify-content-center">
+      <div class="col-5">
+        <%= link_to Settings.sw.url + checkout.catkey, target: '_blank' do %>
+          <%= sul_icon :'sharp-open_in_new-24px' %> View in SearchWorks
+        <% end %>
+      </div>
+      <div class="col-5"></div>
```
cbeer: There must be a better way.... but 🤷‍♂

jkeck: I was thinking the same solution (and similarly unsatisfied).  I have a feeling this `<dl>` column layout will end up changing, and that will probably inform what we can be doing here also.

PR #123, opened 2019-07-15
Add information to requests tab fixes #50

### LINKED ISSUE #50: Requests tab displaying requests
#18 is probably a prerequisite for this ticket (because it will hopefully lay down patterns we can stamp out..). In the meantime, #87 is a ticket to wire up the data.

This ticket is for display requests only; the work to change or cancel requests will be handled in a separate ticket.

Screen Shot 2019-07-10 at 07 55 30

Screen Shot 2019-07-10 at 07 56 17

And inside the expanded request (again, ignoring the form components):

Screen Shot 2019-07-10 at 07 59 58

Note: There's some subtlety around the wait list position that may require a story time to unravel.

mejackreed:
Screen Shot 2019-07-15 at 4 59 47 PM

This PR also generalizes `checkouts.js` -> `listSort.js` so it can be used in multiple places.

### DIFF
--- a/app/controllers/requests_controller.rb
+++ b/app/controllers/requests_controller.rb
@@ -4,8 +4,14 @@
 class RequestsController < ApplicationController
   before_action :authenticate_user!
 
+  # A sufficiently large time used to sort nil values last
+  # TODO: Update before 2099
+  END_OF_DAYS = Time.zone.parse('2099-01-01')
+
   def index
     @response = symphony_client.requests(current_user.patron_key)
-    @requests = @response['fields']['holdRecordList'].map { |request| Request.new(request) }
+    @requests = @response['fields']['holdRecordList'].map { |request| Request.new(request) }.sort_by do |request|
+      [request.pickup_date || END_OF_DAYS, request.fill_by_date || END_OF_DAYS, request.expiration_date || END_OF_DAYS]
+    end
   end
 end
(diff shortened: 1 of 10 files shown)


app/controllers/requests_controller.rb
```diff
@@ -4,8 +4,14 @@
 class RequestsController < ApplicationController
   before_action :authenticate_user!
 
+  # A sufficiently large time used to sort nil values last
+  # TODO: Update before 2099
+  END_OF_DAYS = Time.zone.parse('2099-01-01')
```
camillevilla: 😱😭🔥⚡️🏇🏇🏇🏇

app/views/requests/_request.html.erb
```diff
@@ -0,0 +1,56 @@
+<li class="row d-flex flex-wrap list-group-item">
+  <div class="d-flex flex-row flex-grow-1 justify-content-between w-100 row col-md-4">
+    <div class="w-50 col-md-6 library">
+      <% if request.ready_for_pickup? %>
+        <span class="text-ready font-weight-bold">
+          <%= sul_icon 'sharp-check_circle-24px' %>
+          <span class="d-md-none">Pick up at </span><%= Mylibrary::Application.config.library_map[request.pickup_library] || request.pickup_library %>
```
camillevilla: 🎟 Maybe we can extract this to a helper next sprint @cbeer ?

Issue #120, opened 2019-07-15
Add checkout status-specific detail information


cbeer:
# Grouped by status

## Checked out
- Renewal status
- Borrowed on
- Last Renewed on
- Loan period
- Remaining
- Source

Actions:
- Renew (if eligible)
- View in Searchworks

## Recalled
- Renewal Status
- Borrowed on
- Recalled on
- Source

Actions:
- View in Searchworks

## Overdue
- Renewal status
- Borrowed on
- Fines accruing (if overdue, accruing fines)
- Days overdue
- Source

Actions:
- Renew (if eligible)
- View in Searchworks

## Overdue, billed, assumed lost
- Renewal status
- Borrowed on
- Days overdue
- Assumed lost on
- Assumed lost fee
- Source

Actions:
- Pay $total now --> needs to be reviewed in #188 and #186
- View in Searchworks

# Combined view
- [x] Renewal status
- [x]  Borrowed on
- [x]  Last Renewed on (if renewable)
- [x]  Loan period (if renewable) #142 (logic), UI split out to #204)
- [x]  Remaining (if checkout, not recalled or overdue) #125 
- [x]  Recalled on (if recalled)
- [x]  Days overdue (if overdue) Split out to #185
- [x]  Fines accruing (if overdue, accruing fines (but not assumed lost))  #163
- [ ]  Assumed lost on (if assumed lost) #186 
- [x]  Assumed lost fee (if assumed lost)
- [x] Claimed returned (if claimed returned)
- [x]  Source

Actions:
- [x]  Renew (if eligible and not recalled or assumed lost)
- [x]  View in Searchworks

# TODO
- [ ] display something else in Source if it's from BorrowDirect

camillevilla:
I'm not clear that we need `Days overdue`. It's outlined in the content labels doc but doesn't show up in the mockups. @tallenaz and I worked on some logic that calculates `days_overdue`, but are we actually using this? Symphony returns the `estimatedOverdueAmount` as a dollar amount.

cbeer:
@jvine : ^

jvine:
**Content doc trumps the mockups.** (I created the doc because it was impossible to keep everything accurate and consistent in the mockups.) The doc has had a lot of review & negotiation with @seestone, but is nonetheless open to discussion as we discover what's possible/extra difficult. If this value is problematic, it's fine to set it aside for now. (Also just realized for reserve items we might need hours overdue...)

camillevilla:
[...]

PR #125, opened 2019-07-15
Add days remaining until due date

camillevilla:
Part of #120 

This PR adds logic to calculate days until an item is due. This is added to checkout details, but hidden if the item is overdue.

## Current item ✅
not overdue item

## Overdue ⚠️
overdue item

### DIFF
--- a/app/views/checkouts/_checkout.html.erb
+++ b/app/views/checkouts/_checkout.html.erb
@@ -36,6 +36,10 @@
         <dt class="col-5">Recalled on:</dt>
         <dd class="col-5"><%= l(checkout.recalled_date, format: :short) %></dd>
       <% end %>
+      <% unless checkout.overdue? %>
+        <dt class="col-5">Remaining:</dt>
+        <dd class="col-5"><%= pluralize(checkout.days_remaining, 'day') %> </dd>
+      <% end %>
       <dt class="col-5">Source:</dt>
       <dd class="col-5"><%= Mylibrary::Application.config.library_map[checkout.library] || checkout.library %></dd>
     </dl>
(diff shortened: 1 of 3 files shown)


app/views/checkouts/_checkout.html.erb
```diff
@@ -36,6 +36,10 @@
         <dt class="col-5">Recalled on:</dt>
         <dd class="col-5"><%= l(checkout.recalled_date, format: :short) %></dd>
       <% end %>
+      <% if !checkout.overdue? %>
```
cbeer: Should we use `unless`? Or a new accessor on checkout that means not overdue?

app/views/checkouts/_checkout.html.erb
```diff
@@ -36,6 +36,10 @@
...
+        <dt class="col-5">Remaining:</dt>
+        <dd class="col-5"><%= checkout.days_remaining %> days</dd>
```
cbeer: We should use either:
- i18n's one/other support, or
- [`pluralize`](https://apidock.com/rails/ActionView/Helpers/TextHelper/pluralize)

to make sure this says `1 day` when appropriate

app/views/checkouts/_checkout.html.erb
camillevilla: Went with an `unless` for now.

app/views/checkouts/_checkout.html.erb
camillevilla: Used the `pluralize` approach for now. I'd rather migrate things i18n keys in one PR (soon!) and be thoughtful about how we're organizing them.

Issue #149, opened 2019-07-16
Change my pin for proxy / paid accounts


mejackreed:
Split out of #47 

`/user/patron/changeMyPin`

A separate feature that is needed. After going through reset/request PIN workflow, we need to provide a form that will do the actual changes.

mejackreed:
@jvine I've added design-needed here, but could do something very similar to design in #47 if desired.

jvine:
Okay, here's the desired flow.

Mockup starts here:  https://xd.adobe.com/view/21b404c4-a82b-41fc-6209-e20de8a8edf4-5785/screen/f555ec55-8c27-4fcf-9007-b2324e9d0b51/Reset-PIN

### 1. User requests a PIN
<img width="300" alt="1 request PIN" src="https://user-images.githubusercontent.com/6945691/61337797-a8a43e00-a7eb-11e9-8133-9c7cb9dd9146.png">

### 2. They get a message on the same page
<img width="300" alt="2 PIN requested" src="https://user-images.githubusercontent.com/6945691/61337804-b063e280-a7eb-11e9-9134-dac7afca7807.png">

### 3. They check their email and click the link
Single password field with a show/hide feature strongly preferred over an additional confirm-password field.
<img width="300" alt="3 Email link clicked" src="https://user-images.githubusercontent.com/6945691/61337806-b659c380-a7eb-11e9-904c-f1c9b60f3273.png"><img width="300" alt="3 1 show password" src="https://user-images.githubusercontent.com/6945691/61337807-b954b400-a7eb-11e9-818c-b51f6992611a.png">

### 4. They save their new PIN and can now log in
<img width="300" alt="4 PIN change successful" src="https://user-images.githubusercontent.com/6945691/61337815-bce83b00-a7eb-11e9-8caf-f4a2dbb22799.png">

### 5. Oops, they already used the link or it expired
Note that we're ending up with a different form (different page?) than the successful link. The only useful action they can take at this point is to rerequest a reset link, so we should show them the form where they can do that. 
https://xd.adobe.com/view/21b404c4-a82b-41fc-6209-e20de8a8edf4-5785/screen/bf08fc1d-03c4-475b-b6c0-84ad7ba169ae/Reset-PIN-3

<img width="300" alt="5 link expired" src="https://user-images.githubusercontent.com/6945691/61337825-c2de1c00-a7eb-11e9-96df-544ba48e2ddf.png">

PR #150, opened 2019-07-16
Setup reset/request PIN pages and workflow fixes #47

### LINKED ISSUE #47: Request / reset PIN for proxy or paid accounts
Follow-on to #12.

If the user needs to reset (or set?) their PIN for some reason, they will provide their library ID number. We can use the `user/patron/resetMyPin` API to trigger a reset workflow.

Q: Presumably only non-webauthID users can reset their PINs?
Q: The API docs suggest that the email address might be required to have Symphony send out the email?
Q: Is the result of `resetMyPin` a new PIN, or a link into Symphony to reset the PIN (and, do we want to handle that in MyAccount?)

## Links
From the home page
Screen Shot 2019-07-10 at 07 32 51

From the login page
Screen Shot 2019-07-10 at 07 34 52

## Form:
Screen Shot 2019-07-10 at 07 32 55

## Flash message
After making the symphony API request (regardless of success/failure, assuming e.g. Symphony cares whether the user exists or not)
Screen Shot 2019-07-10 at 07 33 01

mejackreed:
pin

```
Stanford University Libraries PIN Change request
Please follow this link to CHANGE YOUR PIN.

If you have any questions or you are unable to change your PIN, please contact Green Library Privileges at 

Thank you,
Library Privileges Division
Stanford University Libraries
```

### DIFF
--- a/app/services/symphony_client.rb
+++ b/app/services/symphony_client.rb
@@ -73,6 +73,15 @@ class SymphonyClient
     JSON.parse(response.body)
   end
 
+  def reset_pin(library_id)
+    response = request('/user/patron/resetMyPin', method: :post, json: {
+      login: library_id,
+      resetPinUrl: 'future_change_my_pin_path/<RESET_PIN_TOKEN>' # TODO: Based off of #149
+    })
+
+    JSON.parse(response.body)
+  end
+
   private
 
   def authenticated_request(path, headers: {}, **other)
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -26,4 +26,8 @@ class ApplicationController < ActionController::Base
   def authenticate_user!
     redirect_to root_url unless current_user?
   end
+
+  def logout_user!
+    redirect_to logout_path if current_user?
+  end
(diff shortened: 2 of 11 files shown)


mejackreed:
Follow on #149 will complete this workflow

app/services/symphony_client.rb
```diff
@@ -73,6 +73,15 @@ def fines(patron_key)
     JSON.parse(response.body)
   end
 
+  def reset_pin(library_id)
+    response = request('/user/patron/resetMyPin', method: :post, json: {
+      login: library_id,
+      resetPinUrl: 'future_change_my_pin_path/<RESET_PIN_TOKEN>' # TODO: Based off of #149
```
cbeer: Neither here nor there, but I'm not sure I understand what useful task Symphony is doing for us when we call `resetMyPin`.. just.. sending an email?

mejackreed: Yeah exactly.. sending an email with a token tied to the user (that has a short expire) that can then be used to auth for changing the pin

PR #152, opened 2019-07-17
Get borrow limit from symphony and render it in the summaries page 

### LINKED ISSUE #144: Add checkout quota to fee borrower patron info
Split out from #53 

It looks like we can get this information from the `userProfile` policy `chargeLimit` field

jkeck:
Closes #144 

Checkouts

### DIFF
--- a/app/models/patron.rb
+++ b/app/models/patron.rb
@@ -4,6 +4,8 @@
 class Patron
   attr_reader :record
 
+  CHARGE_LIMIT_THRESHOLD = 25_000
+
   PATRON_STANDING = {
     'COLLECTION' => 'Blocked',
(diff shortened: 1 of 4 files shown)


jkeck:
Yeah, I actually named it `charge_limit` at first, but noticed there was a skipped test that wrote the method as `borrow_limit`, so went with that (assuming there was a reason).  Maybe we can chalk it up to the Patron model defines methods in the application's parlance and provides the mapping layer to the underlying API's values (and that may be the seam, or show where the seam may go for an abstraction layer that would facilitate being able to swap out a different ILS in the future).

But yeah, ¯\\_(ツ)_/¯

PR #156, opened 2019-07-17
Adds change PIN form

### LINKED ISSUE #149: Change my pin for proxy / paid accounts

mejackreed:
Fixes #149 and thus finishes the resetPIN / change PIN workflow.

To test this, you must have an account setup that is linked to your email. `JACKPIN` account is currently set to my email. Happy to pair up and show reviewers the whole flow.

resetpin


app/controllers/reset_pins_controller.rb
```diff
@@ -6,9 +6,38 @@ class ResetPinsController < ApplicationController
...
+  private
+
+  def change_pin_with_token_params
+    params.require(%I[token pin])
+  end
+
+  ##
+  # Hacky workaround Rails routing helpers to have the <> not encoded
+  def change_pin_with_token_unencoded_url
+    change_pin_with_token_url('foo').gsub('foo', '<RESET_PIN_TOKEN>')
```
cbeer: Where is `RESET_PIN_TOKEN` coming from? Symphony land?

mejackreed: Yep. Symphony looks for the string `<RESET_PIN_TOKEN>` and injects the actual token in a URL crafted in the email sent to the user.

config/routes.rb
```diff
@@ -21,6 +21,8 @@
 
   get '/reset_pin', to: 'reset_pins#index'
   post '/reset_pin', to: 'reset_pins#reset'
+  get '/change_pin/:token', to: 'reset_pins#change_form', as: :change_pin_with_token
```
mejackreed: Open to other options here on how we craft a URL with a token in it. Tried to keep it simple.

app/views/reset_pins/change_form.html.erb
```diff
@@ -0,0 +1,29 @@
+<h1>Change PIN</h1>
+<%= form_tag change_pin_path, method: :post do %>
+  <%= hidden_field_tag :token, params['token'] %>
```
mejackreed: This comes from the URL that is visited. Symphony when using the `changeMyPin` action, will verify that the correct token came through.

app/services/symphony_client.rb
```diff
@@ -73,15 +73,24 @@ def fines(patron_key)
...
     JSON.parse(response.body)
   end
 
+  def change_pin(token, pin)
+    response = request('/user/patron/changeMyPin', method: :post, json: {
+      resetPinToken: token,
+      newPin: pin
+    })
+
+    response
```
mejackreed: We need the status code so no JSON parsing here.

app/controllers/reset_pins_controller.rb
```diff
@@ -6,9 +6,38 @@ class ResetPinsController < ApplicationController
 
   def index; end
 
+  def change_form; end
```
mejackreed: Names are hard? /shrug open to other suggestions here.

config/routes.rb
mejackreed: Also, we could possibly implement  (future)`change_pin_with_pin` more of a regular "change my password" type workflow

app/views/reset_pins/change_form.html.erb
```diff
@@ -0,0 +1,29 @@
...
+  <div>
+    Your new PIN must:
+    <ul>
+      <li>be 6-25 characters long</li>
```
mejackreed: These rules aren't actually validated anywhere, but its more of a current feature parity thing. We could add future validation if we need to.

app/controllers/reset_pins_controller.rb
cbeer: Got it; would you mind documenting that in the method comment?

mejackreed: Sure thing.. added.

mejackreed:
I'm also 👎 to code climate here.. not really a useful thing. Happy to dismiss, but what's our policy?

Issue #131, opened 2019-07-16
Update checkout status values


cbeer:
In order of priority:

- [x] (red/circle) Recalled
- [x] (red/circle) Recalled $amount (once fines are accruing)
- [x] (no icon) Processing claim (if patron has claimed they returned the item)
- [x] (orange/triangle) Assumed lost $amount (until it’s been paid or returned)
- [x] (orange/triangle) Overdue $amount (once fines are accruing)
- [x] (orange/triangle) Overdue (no fines accruing yet) 
- ~~(no icon) Reserve item (if it is one)~~       
- [x] (no icon) Blank

cbeer:
@saseestone: how do we know if an item is a reserve item?

saseestone:
Just had a conversation with Jennifer about this and we agree that the most interesting thing about this checkout type is not that it's a reserve item...but that it's a short-term loan.  That will be evident from the other details we already have - especially the due date/time.  

So the decision is to not add this to the Checkout detail for now.  We can evaluate later (after user feedback and analytics) if we need it, and then add it in in a future cycle.

I've stricken it from the labels and content doc and put in a note there.

cbeer:
@jvine / @saseestone: what amount should appear for the `Assumed lost $$$`? It seems like our fixture item has an overdue fee, and there are 2 separate fines for the item.

jvine:
After discussion with Sarah:   The amount to show on the Checkouts tab is the overdue fee. The lost item charge(s) are separate, and appear on the Fines tab. At some point we might consider connecting them in some way, but for now: parity.

(Once the user pays the lost item charges, the item is "returned" and the accruing overdue fine is automatically billed and shown on the Fines tab; at that point, the item no longer shows in Checkouts.)

Issue #52, opened 2019-07-10
Summary page and nav bar with dynamic content


cbeer:
The summary page and the nav page display suspiciously similar status data for each of the tabs: 

<img width="380" alt="Screen Shot 2019-07-10 at 08 03 17" src="https://user-images.githubusercontent.com/111218/60980358-2f878100-a2e9-11e9-8943-3885474c6835.png">

<img width="382" alt="Screen Shot 2019-07-10 at 08 02 59" src="https://user-images.githubusercontent.com/111218/60980340-25fe1900-a2e9-11e9-8963-2091d0ca7b66.png">

Hopefully all the necessary information can be retrieved in a single request.

camillevilla:
Closed by #207

PR #168, opened 2019-07-18
Refactor symphony data access to use a single patron query to rule them all

### LINKED ISSUE #169: Refactor Symphony data access patterns
- reduce duplicate requests (or consolidate requests that could be combined into a single request)

cbeer:
Off-hand, it seems 30-50% slower than requests for just a single data type, but also significantly faster than making multiple Symphony API calls.

This should unblock some the work on #131, #120 (which need both checkouts + fines data) and #52 (which needs all the things...).

Fixes #169


mejackreed:
Do we have a use case where we still may want an individual request for some client side / json interaction response thing? Perhaps not now at this moment, but thinking about maybe we need to return only checkout / fine information for something individually?

cbeer:
When you say individually, do you mean just the patron's list of e.g. checkouts, or information about a specific checkout? If the former, I think we're already seeing that the three tabs need data from multiple lists (e.g. checkouts + fines) in order to create the desired behavior. If the latter, maybe; although I suppose we'd retrieve the resource directly instead of diving through the patron resource.

mejackreed:
Yeah I think I was referring the first, but the second also seems plausible. I'm 👍 with this approach here.

cbeer:
Yeah, for the first, I think we have some options.. maybe a method signature like:

`def patron_info(patron_key, include: [:checkouts, :fines, :requests])`

🤷‍♂

PR #174, opened 2019-07-18
Make short term loan's due date show the time instead of just date

### LINKED ISSUE #142: Distinguish between hourly loans and regular checkouts
It looks like this information is encoded in the loanPeriod policy, e.g.

```ruby
{"resource"=>"/policy/loanPeriod",
     "key"=>"3H",
     "fields"=>
      {"policyNumber"=>50,
       "description"=>"0",
       "displayName"=>"3H",
       "periodType"=>{"resource"=>"/policy/periodType", "key"=>"HOURLY"},
       "periodCount"=>3,
       "dayInWeek"=>{"resource"=>"/policy/weekdays", "key"=>"SUNDAY"},
       "timeDue"=>nil,
       "dueDateList"=>[]}}
```

The periodTypes could be:

`["DATELIST", "INDEFINITE", "WEEKLY", "DAILY", "MINUTES", "HOURLY"]`

If the PeriodType is HOURLY or MINUTES, show the due date time instead of just the date:

Screen Shot 2019-07-17 at 13 16 19

jkeck:
Closes #142 

Also updates the Remaining data to use `distance_of_time_in_words` (so we don't get `0 days`)

Screen Shot 2019-07-17 at 4 56 48 PM

### DIFF
--- a/app/helpers/checkouts_helper.rb
+++ b/app/helpers/checkouts_helper.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+# Helper for checkouts views
+module CheckoutsHelper
+  def today_with_time_or_date(date, short_term: false)
+    return l(date, format: :short) unless short_term
+    return l(date, format: :short) unless date.today?
+
+    l(date, format: :time_today)
+  end
+
+  def time_remaining_for_checkout(checkout)
+    return pluralize(checkout.days_remaining, 'day') unless checkout.short_term_loan?
+
+    distance_of_time_in_words(Time.zone.now, checkout.due_date)
+  end
+end
--- a/app/helpers/checkouts_helper.rb
+++ b/app/helpers/checkouts_helper.rb
@@ -2,6 +2,14 @@
 
 # Helper for checkouts views
 module CheckoutsHelper
+  def list_group_item_status_for_checkout(checkout)
+    if checkout.recalled?
+      'list-group-item-danger'
+    elsif checkout.overdue?
+      'list-group-item-warning'
+    end
+  end
+
   def today_with_time_or_date(date, short_term: false)
     return l(date, format: :short) unless short_term
     return l(date, format: :short) unless date.today?
--- a/app/services/symphony_client.rb
+++ b/app/services/symphony_client.rb
@@ -41,6 +41,8 @@ class SymphonyClient
     end
   end
 
+  # Symphony API requires some loooong strings, disabling LineLength for this method
+  # rubocop:disable Metrics/LineLength
   def patron_info(patron_key)
     response = authenticated_request("/user/patron/key/#{patron_key}", params: {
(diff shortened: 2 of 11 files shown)


app/services/symphony_client.rb
```diff
@@ -41,6 +41,17 @@ def session_token
     end
   end
 
+  def checkouts(patron_key)
+    circ_rule_query = 'circulationRule{loanPeriod{periodType{key}}}'
+    item_query = 'item{*,bib{title,author},call{sortCallNumber,dispCallNumber}}'
+
+    response = authenticated_request("/user/patron/key/#{patron_key}", params: {
+      includeFields: "*,circRecordList{*,#{[circ_rule_query, item_query].join(',')}}"
```
cbeer: Can you add these additional fields to the `patron_info` method below (and then remove this method?)

app/views/checkouts/_checkout.html.erb
```diff
@@ -27,7 +27,9 @@
   <div class="collapse w-100" id="collapseExample-<%= checkout.key.parameterize %>">
     <dl class="row justify-content-center">
       <dt class="col-5">Borrowed on:</dt>
```
cbeer: Stupid nitpick for @jvine; if we're showing the time, should this be `Borrowed at` or some such?

app/helpers/checkouts_helper.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+# Helper for checkouts views
+module CheckoutsHelper
```
cbeer: 👏

app/helpers/checkouts_helper.rb
```diff
@@ -0,0 +1,25 @@
...
+    elsif checkout.overdue?
+      'list-group-item-warning'
+    end
+  end
+
+  def today_with_time_or_date(date, short_term: false)
+    return l(date, format: :short) unless short_term
+    return l(date, format: :short) unless date.today?
+
+    "Today at #{date.strftime('%l:%M%P')}"
```
cbeer: Should we add this as another localization format (`time_only` or some such?) 🤷‍♂ ?

jkeck: I embarrassingly didn't realize we could define our own.  lol, good to know.

app/views/checkouts/_checkout.html.erb
jvine: Alternative would be to remove the word "on" globally. So it's just "Borrowed: July 15, 2019" or "Borrowed:  Today at 4:29pm"  

(The "on" has been bugging me anyway.)

cbeer: 👍 ; @jkeck, would you mind doing that as part of this PR?

PR #175, opened 2019-07-18
Update checkout status values for claimed returned

### LINKED ISSUE #131: Update checkout status values

tallenaz:
Fixes #131
Paired with Chris


spec/helpers/application_helper_spec.rb
```diff
@@ -41,4 +41,78 @@
...
+      end
+
+      it 'renders the right html' do
+        expect(content).to have_css('.text-recalled', text: 'Recalled').and(have_css('.sul-icons'))
+      end
+    end
+
+    context 'when an item is claimed returned' do
+      before do
+        allow(checkout).to receive_messages(overdue?: true, lost?: true, claimed_returned?: true, accrued: 666)
```
camillevilla: @tallenaz and @cbeer, just want to confirm my understanding re: `Processing claim` state. 
- I confirmed in #131 and our labels and content doc that it is not supposed to display an icon.
- The item has an orange color because the item is technically still overdue, but no icon is displayed because there's no further action a patron can / should take.

processing claim record with orange background color

Is that right?

camillevilla:
Throwing a screenshot in here for kicks:
## Before
items only displayed as overdue

## After
processing claim, assumed lost states added

spec/helpers/application_helper_spec.rb
cbeer: That's my understanding; in any case, if it's not supposed to be orange, that'll be separate from the status value.

jvine: It is correct as shown.

spec/helpers/application_helper_spec.rb
```diff
@@ -41,4 +41,78 @@
...
+        lost?: false,
+        claimed_returned?: false,
+        accrued: 0
+      )
+    end
+    let(:content) { Capybara.string(helper.render_checkout_status(checkout)) }
+
+    context 'when a recalled item has accrued fines' do
+      before do
+        allow(checkout).to receive_messages(recalled?: true, accrued: 15)
```
jkeck: Is there a reason we're using `allow` here instead of `expect`?

spec/helpers/application_helper_spec.rb
camillevilla: Thanks for the clarification!

spec/helpers/application_helper_spec.rb
tallenaz: I tried setting this `allow` to an `expect` within an `it` block instead of a `before` block, and got a couple of spec failures:

```
 1) ApplicationHelper#render_checkout_status when a recalled item has accrued fines renders the right html
     Failure/Error: expect(content).to have_css('.text-recalled', text: 'Recalled $15').and(have_css('.sul-icons'))
     
     NoMethodError:
       undefined method `css' for nil:NilClass
```

and

```
2) ApplicationHelper#render_checkout_status when a recalled item has accrued fines is recalled with accrued fines
     Failure/Error: expect(checkout).to receive_messages(recalled?: true, accrued: 15)
     
       (InstanceDouble(Checkout) (anonymous)).recalled?(no args)
           expected: 1 time with any arguments
           received: 0 times
     # ./spec/helpers/application_helper_spec.rb:60:in `block (4 levels) in <top (required)>'
```

so it looks like in this case the test object can't receive methods the next expectation requires...

jkeck: Hmm, okay, it seems like if we either allowed ourselves to use `expect` in `before` or multiple `expect`s in a single test we could get around this, but it's not a big deal I suppose.

spec/helpers/application_helper_spec.rb
```diff
@@ -41,4 +41,78 @@
       expect(helper.library_name('NOSUCHLIBRARY')).to eq 'NOSUCHLIBRARY'
     end
   end
+
+  describe '#render_checkout_status' do
```
jkeck: Looks like these tests are duplicated here and in `checkouts_helper_spec`.  Can we remove the one here in `application_helper_spec`?

tallenaz:
@jkeck if you're OK with punting on the allow/expect issue in `checkouts_helper_spec.rb`, I've rebased on the latest `master`, resolving a couple of merge conflicts, and tacked on 1cbce46, since the rebase pulled in enough extra code in the `Checkout` model to alert the cops.

Issue #159, opened 2019-07-17
Display group checkouts tab


camillevilla:
Pre-requisite #143
Under the "Group" tab, a proxy borrower or sponsor account should see the total number of checkouts that belong to the group. Take a look at the `user group holds list` call in our API docs.

![Screen Shot 2019-07-17 at 1 37 15 PM](https://user-images.githubusercontent.com/5402927/61409884-07bf8c80-a898-11e9-83c3-f8abac1331f2.png)

camillevilla:
Remaining tasks after #177:
- [ ] add borrower name to group tab (put at top of expanded list item) for sponsor and proxies
- [ ] styling around tab controls
- [ ] display counts in the tabs

PR #177, opened 2019-07-18
Adds group checkouts, requests, and fines

shelleydoljack:
Part of #159 
Screen Shot 2019-07-18 at 3 37 53 PM
Screen Shot 2019-07-18 at 3 37 34 PM

### Need to:
- [x] write tests
- [~] ~deal with Patron model class length (too long!)~
- [x] make the patron user's tab display `(P=<lastname>)`
- [x] remove the sponsor's data from member_list (to remove it from the group tab)
- [x] remove the current patron's data from the member_list (to remove it from the group tab)

### Split out to another ticket
- [~] add borrower name to group tab (put at top of expanded list item) for sponsor and proxies
- [~] styling around tab controls
- [~] display counts in the tabs (split this out to new ticket probably)

### DIFF
--- a/app/services/symphony_client.rb
+++ b/app/services/symphony_client.rb
@@ -46,7 +46,8 @@ class SymphonyClient
   PATRON_LINKED_RESOURCES = [
     "holdRecordList{*,#{ITEM_RESOURCES}}",
     "circRecordList{*,circulationRule{loanPeriod{periodType{key}},renewFromPeriod},#{ITEM_RESOURCES}}",
-    "blockList{*,#{ITEM_RESOURCES}}"
+    "blockList{*,#{ITEM_RESOURCES}}",
+    'groupSettings{*,responsibility}'
   ].freeze
 
   def patron_info(patron_key)
--- a/app/services/symphony_client.rb
+++ b/app/services/symphony_client.rb
@@ -41,24 +41,27 @@ class SymphonyClient
     end
   end
 
-  # Symphony API requires some loooong strings, disabling LineLength for this method
-  # rubocop:disable Metrics/LineLength
+  ITEM_RESOURCES = 'item{*,bib{title,author},call{sortCallNumber,dispCallNumber}}'
+
+  PATRON_LINKED_RESOURCES = [
+    "holdRecordList{*,#{ITEM_RESOURCES}}",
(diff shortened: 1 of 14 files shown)


shelleydoljack:
The currently logged in user's checkouts and the sponsor's checkouts are not included in the group list:
Screen Shot 2019-07-19 at 5 14 15 PM
Screen Shot 2019-07-19 at 5 14 05 PM

shelleydoljack:
Changes display of proxy borrower name on tab per conversation with Jennifer:
Screen Shot 2019-07-22 at 11 05 11 AM

app/services/symphony_client.rb
```diff
@@ -41,24 +41,28 @@ def session_token
     end
   end
 
-  # Symphony API requires some loooong strings, disabling LineLength for this method
-  # rubocop:disable Metrics/LineLength
+  ITEM_RESOURCES = 'item{*,bib{title,author},call{sortCallNumber,dispCallNumber}}'
+
+  PATRON_LINKED_RESOURCES = [
+    "holdRecordList{*,#{ITEM_RESOURCES}}",
+    "circRecordList{*,circulationRule{loanPeriod{periodType{key}}},#{ITEM_RESOURCES}}",
```
mejackreed: I think that `renewFromPeriod` might have gotten dropped from this. I think this is needed to determine if an item is renewable and is still needed.

app/views/shared/_navigation.html.erb
```diff
@@ -26,3 +26,16 @@
     </li>
   </ul>
 </nav>
+
+<% if patron.group? %>
+<nav class="container">
```
mejackreed: Can we indent the `nav` element underneath the conditional?

spec/controllers/fines_controller_spec.rb
```diff
@@ -42,7 +42,7 @@
       expect(get(:index)).to render_template 'index'
     end
 
-    it 'assigns a list of requests' do
+    it 'assigns a list of fines' do
```
mejackreed: Nice catch

app/services/symphony_client.rb
cbeer: Good catch; I'll add that back.

PR #181, opened 2019-07-19
[WIP] Patron requests are never nil

tallenaz:


app/models/patron.rb
```diff
@@ -94,7 +94,11 @@ def fines
   end
 
   def requests
-    @requests ||= fields['holdRecordList'].map { |request| Request.new(request) }
+    @requests ||= if fields['holdRecordList'].nil?
```
mejackreed: I think Rails also gives us the ability to use `Array.wrap` which could remove our `nil?` check.

@requests ||= Array.wrap(fields['holdRecordList']).map { |request| Request.new(request) }
```suggestion
    @requests ||= Array.wrap(fields['holdRecordList']).map { |request| Request.new(request) }
```

tallenaz:
Closing this, since it looks like the changes @shelleydoljack made to symphony web services made the api respond uniformly across environments and the requests tab works again.

PR #182, opened 2019-07-19
Make test more flexible to handle single and double digit hours

mejackreed:
This was failing the build when run on hours that were single digit `7` but passed for double digit `12`


jkeck:
I think @tallenaz and @cbeer were working on addressing this in their branch as well (so heads up)

mejackreed:
Ah I missed that :( and maybe their solution is probably better? Shall I submit a revert of this?

Issue #130, opened 2019-07-16
Fines details [story]


cbeer:
Ensure the payable fines display the following information:

Above the "fold":
- [x] Reason
- [x] $amount
- [x] Title
- [x] Author

Expand to:
- [x] Billed on: (date)
- [x] Borrowed on: (date)
- [x] Days overdue: count
- [x] Assumed lost on: (date) (if assumed lost)
- [x] (specific type) fee: ($amount) (substitute the `Fine#nice_status` value for specific type.. adding `fee` as needed?)
- [x] Source: (SUL library brief name)

E.g:

<img width="378" alt="Screen Shot 2019-07-17 at 13 51 07" src="https://user-images.githubusercontent.com/111218/61410842-f5465280-a899-11e9-9f3d-64faf2c95572.png">

jkeck:
A few questions about the two remaining pieces for this ticket.

1. `Borrowed on: (date)`
  * Does anybody know how to get this data?  It wasn't immediately obvious how to get, from the `blockList` response to when the particular item that the patron was being charged for checkout date (for this transaction)
2. `Assumed lost on: (date)`
  * What constitutes something that is assumed lost?  I have the current location of the item, however; if the item is replaced, will the current location associated still be accurate?  We could whitelist the status keys that resolve to a `Lost item` label if that would be a more robust solution.
  * Is the date that the item is assumed lost the date that block was created?  Should we get that information from somewhere more explicit than implying from the block creation date?

jvine:
1. From what I learned from Sarah yesterday about this process, I think it's likely not possible to get both the Borrowed and Days overdue amounts for Assumed lost items. The lost item fee is a separate record from the original checkout. I think these two values may be available only for Overdue fees. 

2. Looking at the Labels & content doc - the "Assumed lost on (date)" has been removed because it is the same as the "Billed" date. 

Assumed lost goes like this:
Item is very overdue on the checkouts tab. It passes a threshold. Someone says "bill this one", and _separate_ bill records are created for the lost item fee and processing fee. (The lost item fee is high and will block the user.) The overdue item on the checkouts tab remains there, now marked Assumed lost. The user is still responsible for the overdue fine - the lost item fee does not replace it. But they can't pay the overdue until they either:
a) return the item (at which point the lost item fee is cleared; but the user still has to pay the processing fee - and now the overdue fee)
b) pay the lost item fee, which essentially returns the book. Then same, they have to pay the overdue fee.

Lesson: don't lose books, kids.

PR #183, opened 2019-07-19
Add more details to the Fines page

jkeck:
Part of #130 

This PR adds `Days overdue` and the `{Type} Fee` data.

Screen Shot 2019-07-19 at 8 40 05 AM

### DIFF
--- a/app/helpers/fines_helper.rb
+++ b/app/helpers/fines_helper.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+# Helper module for Fines
+module FinesHelper
+  def nice_status_fee_label(status)
+    return status if status.ends_with?('fee')
+
+    "#{status} fee"
+  end
+end
--- a/app/models/fine.rb
+++ b/app/models/fine.rb
@@ -88,6 +88,10 @@ class Fine
     fields['owed']['amount'].to_d
   end
 
+  def fee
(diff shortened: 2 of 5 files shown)


app/helpers/fines_helper.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+# Helper module for Fines
+module FinesHelper
+  def nice_status_fee_label(status)
+    return status if status.ends_with?('fee')
+
+    "#{status} fee"
```
jkeck: I thought about just adding this to the model, and would be happy to do that.

mejackreed: I'm fine with it here.. seems like it fits the cognitive model of where to place something like this.

app/models/fine.rb
```diff
@@ -77,10 +77,21 @@ def bill_date
     Time.zone.parse(fields['billDate']) if fields['billDate']
   end
 
+  def days_overdue
+    return unless bill_date
+    return unless bill_date.past?
+
+    ((Time.zone.now - bill_date).to_i / 60 / 60 / 24).round + 1
```
jkeck: I think this (+ 1) is the correct logic as illustrated by the test. I believe if the bill_date is today's date, the item is already 1 day overdue (and not 0 days overdue)?

PR #208, opened 2019-07-23
Adds payment history using symphony legacy ws client

jgreben:
Full view:

Screen Shot 2019-07-23 at 9 09 40 AM

Responsive view:

Screen Shot 2019-07-23 at 9 10 02 AM

Lists item showing truncation with ellipses after two lines:

Screen Shot 2019-07-24 at 8 57 40 AM

Screen Shot 2019-07-24 at 8 57 24 AM

### DIFF
--- a/app/assets/stylesheets/modules/record.scss
+++ b/app/assets/stylesheets/modules/record.scss
@@ -1,6 +1,14 @@
+
 .record-title {
   @include h3-index-title-font;
   @include h3-index-title-small-font;
+  -webkit-box-orient: vertical;
+  box-orient: vertical;
+  display: -webkit-box;
+  -webkit-line-clamp: 2;
+  line-clamp: 2;
+  overflow: hidden;
+  text-overflow: ellipsis;
 }
 
 .list-header {
--- a/app/controllers/fines_controller.rb
+++ b/app/controllers/fines_controller.rb
@@ -5,15 +5,38 @@ class FinesController < ApplicationController
   before_action :authenticate_user!
 
   def index
-    @fines = if params[:group]
-               patron.group_fines
-             else
-               patron.fines
-             end
-    @checkouts = if params[:group]
-                   patron.group_checkouts.sort_by(&:due_date)
-                 else
-                   patron.checkouts.sort_by(&:due_date)
-                 end
+    @fines = fines
+    @checkouts = checkouts
+    @payments = if payments
+                  Array.wrap(payments).map { |payment| Payment.new(payment) }
+                else
+                  []
+                end
+  end
+
+  private
+
+  def fines
+    if params[:group]
+      patron.group_fines
+    else
+      patron.fines
+    end
+  end
+
+  def checkouts
+    if params[:group]
+      patron.group_checkouts.sort_by(&:due_date)
+    else
+      patron.checkouts.sort_by(&:due_date)
+    end
+  end
+
+  def payments_response
+    symphony_legacy_client.payments(symphony_client.session_token, patron)
+  end
+
+  def payments
+    (Hash.from_xml(Nokogiri::XML(payments_response).to_s) || {}).dig('LookupPatronInfoResponse', 'feeInfo')
   end
 end
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -1,0 +1,65 @@
+# frozen_string_literal: true
+
+# Model for a Payment on the Fine page
+class Payment
+  attr_reader :record
+
+  def initialize(record)
+    @record = record
+  end
+
+  def key
+    record['billNumber']
+  end
+
+  def bill_description
+    record['billReasonDescription']
+  end
+
+  def bill_amount
+    record['amount']
+  end
+
+  def bill_date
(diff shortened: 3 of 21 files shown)


app/models/payment.rb
```diff
@@ -0,0 +1,66 @@
...
+  def resolution
+    fee_pay_info['paymentTypeDescription'] || payment_amount && 'Paid' || 'Unknown'
+  end
+
+  def paid_fee
+    resolution.match(/credit or debit card/) || false
+  end
+
+  def paid_fee?
+    paid_fee.to_s.eql?('credit or debit card')
```
jkeck: It looks like `paid_fee` is only consumed by this `paid_fee?`.  It seems a bit odd that we maybe casting `false` to a string and then doing an `eql?` check.  Could we just use the existence of a match data to determine if it's a paid fee or not?

app/models/payment.rb
```diff
@@ -0,0 +1,66 @@
...
+  def payment_amount
+    fee_pay_info && fee_pay_info['paymentAmount']
+  end
+
+  def payment_date
+    fee_pay_info && Time.strptime(fee_pay_info['paymentDate'], '%Y-%m-%d')
+  end
+
+  def resolution
+    fee_pay_info['paymentTypeDescription'] || payment_amount && 'Paid' || 'Unknown'
```
jkeck: This `||` `&&` `||` is a bit confusing and doesn't appear to have all of its various logic paths tested (which might have made it a bit more clear).  Maybe this could use something like explicit returns when various criteria is met and then fall back to the default value.  Would be happy to chat w/ you about what I'm talking about if it isn't clear.

app/controllers/fines_controller.rb
```diff
@@ -7,5 +7,8 @@ class FinesController < ApplicationController
   def index
     @fines = patron.fines
     @checkouts = patron.checkouts
+    @payments_response = symphony_legacy_client.payments(symphony_client.session_token, patron)
+    payments = Hash.from_xml(Nokogiri::XML(@payments_response).to_s)['LookupPatronInfoResponse']['feeInfo']
+    @payments = payments ? payments.map { |payment| Payment.new(payment) } : {}
```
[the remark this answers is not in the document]
jgreben: `Hash.from_xml(Nokogiri::XML(@payments_response).to_s)` is a hash that has `feeInfo` which is an array within that hash. So `payments` is an array

app/assets/stylesheets/modules/record.scss
```diff
@@ -1,3 +1,11 @@
+.payment-title {
+  -webkit-box-orient: vertical;
+  display: -webkit-box;
+  -webkit-line-clamp: 2;
```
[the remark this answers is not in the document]
jgreben: https://caniuse.com/#search=line-clamp shows older browsers, IE11, and some Android platforms do not support line-clamp. If it is not supported in browser we will just see a long title.

app/controllers/fines_controller.rb
```diff
@@ -5,15 +5,36 @@ class FinesController < ApplicationController
...
-                   patron.group_checkouts.sort_by(&:due_date)
-                 else
-                   patron.checkouts.sort_by(&:due_date)
-                 end
+    @fines = fines
+    @checkouts = checkouts
+    @payments = if payments
+                  Array.wrap(payments).map { |payment| Payment.new(payment) }
+                else
+                  {}
```
jkeck: So now that we're using `Array.wrap` and ensuring that we're returning an array, we should make the fallback return object an array too (instead of a hash here)

app/controllers/fines_controller.rb
```diff
@@ -5,15 +5,36 @@ class FinesController < ApplicationController
...
+                end
+  end
+
+  def fines
```
jkeck: We may want to consider making this `private`

spec/support/fake_symphony.rb
```diff
@@ -32,6 +32,16 @@ class FakeSymphony < Sinatra::Base
     json_response 200, "patron/#{params[:key]}.json"
   end
 
+  get '/symws/rest/patron/lookupPatronInfo' do
+    content_type :'text/html'
+    status 200
+    begin
+      File.open(File.dirname(__FILE__) + '/fixtures/patron/' + "#{params['userID']}.xml").read
```
jkeck: Awesome that you figured this out.  I wonder if we should put these files under a separate directory (like `/fixtures/patron/payment_history` or something like that so that we can distinguish these files from the files that are meant to represent a web services response.

app/services/symphony_legacy_client.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+# HTTP client wrapper for making requests to Symws
+class SymphonyLegacyClient
+  def payments(session_token, patron)
+    response = request('/rest/patron/lookupPatronInfo', params: {
+      clientID: 'SymWSTestClient',
```
shelleydoljack: Should this be a shared config?

jgreben:
> Looks good to me. I surmise that past group payments will get added in another PR.

Indeed but that is something that even the legacy web services did not include :(

PR #209, opened 2019-07-23
Canceling a single request

mejackreed:
- [x] Cancel a single request that is ready for pickup
 - [x] Change a pickup library that is not yet ready for pickup
 - [x] Change a not needed by date that is not yet ready for pickup
 - [x]  Figure out the correct api call for not needed by
 - [x] Figure out how to put what libraries can pickup be changed to
 - [x] Cancel a request that is not yet ready for pickup
 - [x] ~Put it in a modal~ Split to #242

change

### DIFF
--- a/app/controllers/requests_controller.rb
+++ b/app/controllers/requests_controller.rb
@@ -13,4 +13,21 @@ class RequestsController < ApplicationController
       [request.expiration_date || END_OF_DAYS, request.fill_by_date || END_OF_DAYS]
     end
   end
+
+  def destroy
+    @response = symphony_client.cancel_hold(*cancel_hold_params)
+    case @response.status
+    when 200
+      flash[:success] = t 'mylibrary.request.cancel.success_html', title: params['title']
+    else
+      flash[:error] = t 'mylibrary.request.cancel.error_html', title: params['title']
+    end
+    redirect_to requests_path
+  end
+
+  private
+
+  def cancel_hold_params
+    params.require(%I[resource id])
+  end
 end
--- a/app/controllers/requests_controller.rb
+++ b/app/controllers/requests_controller.rb
@@ -15,7 +15,7 @@ class RequestsController < ApplicationController
   end
 
   def edit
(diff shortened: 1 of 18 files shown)


app/controllers/requests_controller.rb
```diff
@@ -19,4 +19,65 @@ def index
                   end
                 end
   end
+
+  def edit
+    @request = patron.requests.select { |r| r.key == params['id'] }.first
```
cbeer:
```suggestion
    @request = patron.requests.find { |r| r.key == params['id'] }
```

app/models/library_location.rb
```diff
@@ -0,0 +1,42 @@
...
+#  Class to handle configuration and logic around library codes and labels
+#  Adapted from https://github.com/sul-dlss/sul-requests/blob/master/app/models/library_location.rb
+###
+class LibraryLocation
+  def initialize(home_location)
+    @home_location = home_location
+  end
+
+  def additional_pickup_libraries(current_pickup_library)
+    pickup_libraries.reject { |l| l == current_pickup_library }
```
cbeer:
```suggestion
    pickup_libraries.except(current_pickup_library)
```

app/models/library_location.rb
```diff
@@ -0,0 +1,42 @@
...
+  # rubocop:enable Style/EmptyCaseCondition
+
+  private
+
+  def library_specific_pickup_libraries?
+    config.library_specific_pickup_libraries.key?(@home_location)
+  end
+
+  def location_specific_pickup_libraries?
+    config.location_specific_pickup_libraries.key?(@home_location)
```
cbeer: Maybe we should pull `config.location_specific_pickup_libraries` out to a private method so it's a little less unwieldy.

mejackreed: Great recommendation!

mejackreed:
@camillevilla oops looks like I had that commented out. I popped on 13d0f1d which should resolve the issue. I had it commented out before I knew about the morison way to create requests.

PR #222, opened 2019-07-23
 Add privileges expire information to the summary; fixes #216

### LINKED ISSUE #216: Add patron privilege expiration date to the summary screen
Consult [labels & content](https://docs.google.com/document/d/103F_NJJsBx-Bb7Op6nHGmwqylVcl5EvI7V67-PLa4Ho/edit) doc
This applies to fee borrowers and proxy accounts

cbeer:
Screen Shot 2019-07-23 at 15 56 37

I'm assuming the privileges expiration date is only set in Symphony for the users we want to show this information.

### DIFF
--- a/app/views/summaries/index.html.erb
+++ b/app/views/summaries/index.html.erb
@@ -8,16 +8,16 @@
   <h2><%= "#{@patron.first_name} #{@patron.last_name}" %></h2>
   <dl class="row">
     <% if @patron.patron_type == 'Fee borrower' %>
-      <dt class='col-2 col-sm-1 font-weight-normal'>Type</dt>
+      <dt class='col-2 col-sm-1 font-weight-normal'>Type:</dt>
       <dd class='col-10 col-sm-11 patron-type'><%= @patron.patron_type %></dd>
     <% end %>
-    <dt class="col-2 col-sm-1 font-weight-normal">Status</dt>
+    <dt class="col-2 col-sm-1 font-weight-normal">Status:</dt>
     <dd class="col-10 col-sm-11 patron-status"><%= @patron.status %></dd>
     <% if @patron.borrow_limit %>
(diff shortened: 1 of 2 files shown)


jvine:
> I'm assuming the privileges expiration date is only set in Symphony for the users we want to show this information.

If I'm understanding this correctly,  no.  Students (for example) have expiration dates, but we don't want to display it to them. We don't intend to expose the expiration date for any SUNet ID borrowers - only for PIN borrowers.

app/views/summaries/index.html.erb
```diff
@@ -8,16 +8,20 @@
...
-    <dt class="col-2 col-sm-1 font-weight-normal">Status</dt>
+    <dt class="col-2 col-sm-1 font-weight-normal">Status:</dt>
     <dd class="col-10 col-sm-11 patron-status"><%= @patron.status %></dd>
     <% if @patron.borrow_limit %>
-      <dt class="col-2 col-sm-1 font-weight-normal">Borrower limit</dt>
+      <dt class="col-2 col-sm-1 font-weight-normal">Borrower limit:</dt>
       <dd class="col-10 col-sm-11 patron-status"><%= pluralize(@patron.borrow_limit, 'checkout') %></dd>
     <% end %>
-    <dt class="col-2 col-sm-1 font-weight-normal">Email</dt>
+    <% if @patron.expired_date %>
```
camillevilla: Ditto @jvine https://github.com/sul-dlss/mylibrary/pull/222#issuecomment-514417024
I remember talking to @shelleydoljack and @saseestone about this. It seems like there is more turnover in student priv expiration dates due to enrollment issues, so displaying the expiration date for student users causes unnecessary headaches. 

Could you add an explicit check here that a patron is either a fee borrower or a proxy borrower?

PR #223, opened 2019-07-23
Adds borrower name in the group content.

shelleydoljack:
Part of #159 
Displays borrower name on group tab:
Screen Shot 2019-07-23 at 10 58 56 PM
Does not display borrower name for self:
Screen Shot 2019-07-23 at 10 58 38 PM

### DIFF
--- a/app/models/patron.rb
+++ b/app/models/patron.rb
@@ -94,6 +94,16 @@ class Patron
     end
   end
 
+  def member_list_names
+    @member_list_names ||= begin
+      member_list.each_with_object({}) { |member, hash| hash[member.key] = member.first_name }
+    end
+  end
+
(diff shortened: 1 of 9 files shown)


app/models/patron.rb
```diff
@@ -94,6 +94,16 @@ def member_list
     end
   end
 
+  def member_list_names
```
cbeer: I think something like this might be a more idiomatic way to write this (plus, with memoization)

```
@member_list_names ||= begin
  member_list.each_with_object({}) { |hash, member| hash[member.key] = member.first_name }
end
```

shelleydoljack: Thanks for the suggestion!

PR #232, opened 2019-07-24
Add item icons

cbeer:


cbeer:
@jvine : The item icons from #172 don't seem to match up well with the mapping from Searchworks. I've mapped things where similar names existed, but I wonder if we're still missing some icons?

Check out 7f759a5 for the changes.

PR #238, opened 2019-07-24
Truncate authors to one line

### LINKED ISSUE #217: Truncate item titles and authors
Show:
- 2 lines of a title
- 1 line of an author

in checkouts, requests, and fines.

jkeck:
Closes #217 

Note that the call number is two lines already, so we have space for two lines of the author name if needed.

Screen Shot 2019-07-24 at 1 21 40 PM

### DIFF
--- a/app/assets/stylesheets/modules/record.scss
+++ b/app/assets/stylesheets/modules/record.scss
@@ -5,6 +5,10 @@
   @include truncate-by-lines(2);
 }
 
+.record-author {
+  @include truncate-by-lines(1);
+}
+
 .list-header {
   margin-left: 0;
   margin-right: 0;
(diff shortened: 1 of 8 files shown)


app/assets/stylesheets/modules/record.scss
```diff
@@ -2,13 +2,11 @@
 .record-title {
```
cbeer: I'm 👍 on this, with a lingering doubt that maybe we should call these classees `.clamp-1` and `.clamp-2` to maybe be more bootstrap-y?

jkeck: Yeah, had the same though, now that two of us have had the same thought, maybe it's a good idea.  I'll make the change.

jkeck: And that is done now.

PR #247, opened 2019-07-25
Navigation tabs should 'remember' if you're in a group context or not

cbeer:
After selecting a groups tab, navigation links should keep the user in the group context.


cbeer:
@jvine , can you confirm this is the expected behavior?

jvine:
Yes, this is correct.

PR #248, opened 2019-07-25
Add unseenRenewablesCheck as part of Checkout renewable? logic

### LINKED ISSUE #237: Add check for number of unseen renewals remaining = 0
@mejackreed ran into this issue when trying to renew an item in demo. 
Once a user has 0 `unseen renewals remaining`, they cannot renew the item and must bring it to the desk. Let's add a more specific error message than "Sorry! Something went and {title} was not renewed."

Screen Shot 2019-07-24 at 11 59 16 AM

- [x] confirm whether we check unseen renewals remaining. 
- [x] If not, add a method that we can use for picking an error message

Current My Account error message is: "Only {x number of renewals} allowed for this item"

mejackreed:
fixes #237

### DIFF
--- a/app/models/checkout.rb
+++ b/app/models/checkout.rb
@@ -55,7 +55,7 @@ class Checkout
   ##
   # Is this item renewable
   def renewable?
-    unseen_renewals_remaining.positive? && Time.zone.now > renewable_at if renewable_at
+    unseen_renewals_remaining.positive? && renewable_at.past? if renewable_at
   end
 
   def unseen_renewals_remaining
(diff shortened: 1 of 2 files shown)


app/models/checkout.rb
```diff
@@ -55,7 +55,11 @@ def renewal_date
   ##
   # Is this item renewable
   def renewable?
-    Time.zone.now > renewable_at if renewable_at
+    unseen_renewals_remaining.positive? && Time.zone.now > renewable_at if renewable_at
```
cbeer: I think there's a Rails-y way to write this check

PR #252, opened 2019-07-25
Add eResource access message for library_id logged in users, fixes #214

### LINKED ISSUE #214: patron info: eResource access
Split out from #53 
- blocked until we have LDAP permissions
- This only displays for fee borrowers

mejackreed:
This could have been implemented using the existing `shibboleth?` method, but would not work in local development. I've implemented the `library_id_user` to reduce developer confusion.

Screen Shot 2019-07-25 at 2 30 57 PM

We may also want to adjust markup here (column lengths) but thought that might want to be deferred once we get all of the information needed here.

### DIFF
--- a/app/views/patron/_patron.html.erb
+++ b/app/views/patron/_patron.html.erb
@@ -12,7 +12,7 @@
   <% end %>
   <% if patron.fee_borrower? %>
     <dt class="col-4 col-md-2 font-weight-normal">eResource access:</dt>
-    <dd class="col-10 col-sm-11 patron-status">In-library only</dd>
+    <dd class="col-8 col-md-10 patron-status">In-library only</dd>
   <% end %>
   <% if patron.expired_date && (patron.fee_borrower? || patron.proxy_borrower?) %>
     <dt class="col-2 col-sm-1 font-weight-normal">Privileges expire:</dt>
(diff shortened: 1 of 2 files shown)


app/views/patron/_patron.html.erb
```diff
@@ -10,6 +10,10 @@
     <dt class="col-2 col-sm-1 font-weight-normal">Borrower limit:</dt>
     <dd class="col-10 col-sm-11 patron-status"><%= pluralize(patron.borrow_limit, 'checkout') %></dd>
   <% end %>
+  <% if current_user.library_id_user? %>
```
cbeer: Per @saseestone: we should tie this to `patron.fee_borrower?`, not how they logged in.

mejackreed: That simplifies things :) ! 

I've pushed up the change and removed the extra logic I added.

app/views/patron/_patron.html.erb
```diff
@@ -10,6 +10,10 @@
     <dt class="col-2 col-sm-1 font-weight-normal">Borrower limit:</dt>
     <dd class="col-10 col-sm-11 patron-status"><%= pluralize(patron.borrow_limit, 'checkout') %></dd>
   <% end %>
+  <% if patron.fee_borrower? %>
+    <dt class="col-2 col-sm-1 font-weight-normal">eResource access:</dt>
```
camillevilla:
```suggestion
    <dt class="col-4 col-md-2 font-weight-normal">eResource access:</dt>
```

app/views/patron/_patron.html.erb
```diff
@@ -10,6 +10,10 @@
...
+    <dd class="col-10 col-sm-11 patron-status">In-library only</dd>
```
camillevilla:
```suggestion
    <dd class="col-8 col-md-10 patron-status">In-library only</dd>
```

PR #257, opened 2019-07-25
Payment workflow

jgreben:
Payment amount with payment button:

Screen Shot 2019-07-25 at 2 16 29 PM

After going to CyberSource and clicking Cancel button:

Screen Shot 2019-07-25 at 2 17 38 PM

After successful payment and clicking Return to Website:

Screen Shot 2019-07-25 at 5 04 53 PM

After a moment when they navigate away from the Fines tab or click on the Fines tab again, Symphony will have processed the payment, the payment will appear in the payment history, and the payment pending status will disappear.

### DIFF
--- a/spec/controllers/fines_controller_spec.rb
+++ b/spec/controllers/fines_controller_spec.rb
@@ -3,7 +3,13 @@
 require 'rails_helper'
 
 RSpec.describe FinesController do
-  let(:mock_patron) { instance_double(Patron) }
+  let(:mock_patron) { instance_double(Patron, group?: false, barcode: '1234') }
+
+  let(:fines) do
+    [
+      instance_double(Fine, key: '1', sequence: '1', owed: '12', status: 'BADCHECK')
+    ]
+  end
 
   let(:mock_legacy_client) do
     instance_double(
@@ -39,12 +45,6 @@ RSpec.describe FinesController do
   context 'with an authenticated request' do
     let(:user) do
       { username: 'somesunetid', patron_key: '123' }
-    end
-
-    let(:fines) do
-      [
-        instance_double(Fine, key: '1')
-      ]
     end
 
     let(:checkouts) do
@@ -109,12 +109,6 @@ RSpec.describe FinesController do
   context 'with an authenticated request for group fines' do
     let(:user) do
       { username: 'somesunetid', patron_key: '123' }
-    end
-
-    let(:fines) do
-      [
-        instance_double(Fine, key: '1')
-      ]
     end
 
     let(:checkouts) do
--- a/app/controllers/payments_controller.rb
+++ b/app/controllers/payments_controller.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+# Controller for Payments from CyberSource
+class PaymentsController < ApplicationController
+  # CyberSource is posting back to our controller
+  skip_forgery_protection
+
+  def accept
+    redirect_to fines_path(payment_pending: true), flash: {
+      success: (t 'mylibrary.fine_payment.success_html', amount: params[:req_amount])
+    }
+  end
+
+  def cancel
+    redirect_to fines_path, flash: { error: 'Payment canceled. No payment was made for the payable amount.' }
+  end
+end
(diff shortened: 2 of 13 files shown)


app/assets/stylesheets/base.scss
```diff
@@ -9,6 +9,12 @@ a {
     fill: $sul-link-color;
   }
 
+  &.btn {
```
cbeer: If you haven't, I'd try rebasing and see if this is still necessary. There was a fix recently that should auto-calculate the right icon color for a given button type.

app/controllers/fines_controller.rb
```diff
@@ -7,13 +7,26 @@ class FinesController < ApplicationController
   def index
     @fines = fines
     @checkouts = checkouts
+    @pay_link = cybersource
```
cbeer: It feels a little funny to me to make this an instance variable on the controller; is it maybe more of a helper method?

app/controllers/fines_controller.rb
```diff
@@ -31,4 +44,32 @@ def payments_response
...
+      reason: fines.map(&:status).join(','),
+      billseq: fines.map(&:sequence).values_at(0, -1).join('-'),
+      amount: format('%.2f', fines.sum(&:owed))
+    }
+  end
+
+  def user_params
+    {
+      session_id: symphony_client.session_token,
+      group: patron.group? && 'G',
```
cbeer: This is forcing a user with a group to always pay a group fine? It seems like it should probably be contextual (based on the `params[:group]` value)?

app/controllers/fines_controller.rb
```diff
@@ -31,4 +44,32 @@ def payments_response
...
+    {
+      reason: fines.map(&:status).join(','),
+      billseq: fines.map(&:sequence).values_at(0, -1).join('-'),
+      amount: format('%.2f', fines.sum(&:owed))
+    }
+  end
+
+  def user_params
+    {
+      session_id: symphony_client.session_token,
```
cbeer: This is exposed to the user? We probably don't want to leak our session token to them or they'll be able to connect to symphony as our user, right?

config/environments/development.rb
```diff
@@ -29,6 +29,10 @@
     config.cache_store = :null_store
   end
 
+  # Disable request forgery protection in dev environment.
```
cbeer: I'm a little confused by this (and, I guess, I worry it isn't just limited to local development); it seems like it might be worth looking into a little more to see if there's a better way to handle this without globally disabling such an important security feature.

app/assets/stylesheets/base.scss
jgreben: Removed the extra CSS.

app/controllers/fines_controller.rb
jgreben: Creates a payments_helper module taht is called from the view.

app/controllers/fines_controller.rb
jgreben: Now passing in a `group_choice` param to the `payments_helper`.

app/controllers/fines_controller.rb
jgreben: Now just using `SecureRandom.uuid`. The session id used to be the internal symphony Webcat session but can be anything and just has to exist unless we change the `payment_form` code.

config/environments/development.rb
jgreben: I added this to the controller instead:

`skip_before_action :verify_authenticity_token, only: :create, unless: :production?`

spec/controllers/fines_controller_spec.rb
```diff
@@ -43,7 +43,7 @@
 
     let(:fines) do
```
tallenaz: This `fines` variable is identical to the variable on line 116, can we dry it up?

app/controllers/fines_controller.rb
jgreben: Moved this to new PaymentsController

app/controllers/fines_controller.rb
jgreben: Moved this to helper method

app/controllers/fines_controller.rb
jgreben: Now using just a string 'UNUSED'

config/environments/development.rb
jgreben: Moved this to PaymentController

PR #274, opened 2019-07-26
Put the top navbar elements inside a container

### LINKED ISSUE #253: Fix left alignments of all elements
Screen Shot 2019-07-25 at 14 21 57

These should all be aligned to the left, not ragged as shown (logo left position is the correct one)

cbeer:
Screen Shot 2019-07-26 at 08 36 18

Fixes #253, maybe?


app/views/shared/_top_navbar.html.erb
```diff
@@ -1,19 +1,21 @@
 <nav id="topnav" class="navbar navbar-expand-always container" role="navigation">
```
mejackreed: So a double `container`?

cbeer: Thanks for the catch; no, not a double one, just putting one where bootstrap expects it to be.

Issue #260, opened 2019-07-26
Add additional Google Analytics events(?) for user interactions


cbeer:
- [x] Pay fines
- [x] Renew individual items
- [x] Renew all items
- [x] Cancel single request
- [x] Change single requests
- [x] ~Change all requests~ Canceled in https://github.com/sul-dlss/mylibrary/issues/54
- [x] Expand/collapse details
- [x] Following link out to searchworks
- [x] Contact access services (or library) form interactions
- [x] Sort

mejackreed:
What should we be tracking for the "Contact" event? When the form is submitted? What library was contacted?

mejackreed:
Also, with "Change single requests" do we just want to capture that they pressed that button? Or more information?

mejackreed:
Same question for "pay fine"?

jvine:
Contact -  the "To" value (specific library, or circ&privs)

Change single request - which action did they do:  change date, change library, cancel. (What if they change both date and library? Is that two separate events or one?) (And what if they change date and library and click cancel, which would indicate a UI problem?)

Pay fine - Since the payment can represent several fees and fines, probably just the fact that a payment was made. Amount seems iffy - like maybe this is not the place to be capturing how much money was collected in the past week... (@saseestone?)

cbeer:
For pay fine, it seems like it'd also be nice to know:

- a) they clicked a button to go out to Cybersource
- b) they paid their fine or hit the cancel button

saseestone:
☝️ This is definitely not the place to record how much money was collected.  😄 

Agreed that collecting that a payment was made seems good.

jvine:
Ah yes, paid vs cancelled.

mejackreed:
@cbeer I'm not sure how to track that complex interaction with our current analytics tools. Do you have any ideas how to be able to start tracking that round trip transaction?

PR #275, opened 2019-07-26
Start adding custom analytics events

mejackreed:
Part of #260 , some of those items are blocked by inflight work.


mejackreed:
CodeClimate is bogus here.

PR #289, opened 2019-07-26
Add initial support for displaying active BorrowDirect requests in requests tab

### LINKED ISSUE #212: Add support for displaying BorrowDirect requests
In the requests app we use the [`borrow_direct`](https://rubygems.org/gems/borrow_direct) gem.

That gem [provides the ability](https://github.com/jrochkind/borrow_direct#get-patrons-current-requests-requestquery-api) to get a users requests given their patron barcode (which should allow us to list them in the checkouts page).

jkeck:
Closes #212 

## Mobile
Screen Shot 2019-07-26 at 4 40 47 PM

## Desktop
Screen Shot 2019-07-26 at 4 38 19 PM

### DIFF
--- a/app/models/borrow_direct_requests.rb
+++ b/app/models/borrow_direct_requests.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+##
+# Wrap the BorrowDirect::RequestQuery in a class that we
+# can inject our patron barcode from (and do error handling)
+class BorrowDirectRequests
+  attr_reader :patron
+  def initialize(patron)
+    @patron = patron
+  end
+
+  def requests
+    request_client.requests('open', true).map { |request| BorrowDirectRequests::Request.new(request) }.select(&:active?)
+  rescue BorrowDirect::Error
+    []
+  end
+
+  private
+
+  def request_client
+    @request_client ||= BorrowDirect::RequestQuery.new(patron.barcode)
+  end
+
+  ##
+  # Wrap the BorrowDirect::RequestQuery::Item in a class
+  # so we can give it a similar iterface to Symphony Requests
+  class Request < SimpleDelegator
+    # Request becomes ON_LOAN once we receive it (and should show as a request ready for pickup/checkout)
+    # Request becomes COMPLETED once the uesr returns it
+    ACTIVE_REQUEST_STATUSES = %w[
+      ENTERED IN_PROCESS SHIPPED
+    ].freeze
+
+    def active?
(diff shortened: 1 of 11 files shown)


jkeck:
FWIW, I don't believe test users can have BorrowDirect requests (and they don't have a real test system that we can use), so we kind of have to use library IDs in order to test this.  I would be happy to help anybody review w/ some library IDs (or @saseestone can too)

app/models/borrow_direct_requests.rb
```diff
@@ -0,0 +1,54 @@
...
+# can inject our patron barcode from (and do error handling)
+class BorrowDirectRequests
+  attr_reader :patron
+  def initialize(patron)
+    @patron = patron
+  end
+
+  def requests
+    request_client.requests('open', true).map { |request| BorrowDirectRequests::Request.new(request) }.select(&:active?)
+  rescue BorrowDirect::Error
```
cbeer: Should we report these errors to honeybadger, or are they just expected?

app/models/patron.rb
```diff
@@ -151,6 +151,16 @@ def to_partial_path
 
   private
 
+  def borrow_direct_requests
+    return [] if proxy_borrower? # Proxies can't submit borrow direct requests, so don't check.
```
cbeer: Just confirming that fee borrowers can do BD?

jkeck: I would guess probably not.  @saseestone, can you confirm?

app/models/borrow_direct_requests.rb
jkeck: Unfortunately expected.  They return an error when the user has no requests (`PUBQR004: No result for query`) and always in any non-production mode since their default server always returns an error (`PUBSC003: Invalid parameter - Invalid PartnershipId`), which I presume is because they never added us to their test system.

The BorrowDirect gem doesn't make any distinction between these and returns the top level `BorrowDirect::Error`.  The gem does have some facility for setting expected error codes (although does not include this case), but the RequestQuery class has some assumptions about what is returned from the API and doesn't handle this case well.  I would be happy to submit a PR upstream to address that, which would allow us to report the real exceptions (if folks that it would be worth doing)

cbeer: 👍 ; I'm happy to punt for now.

app/models/patron.rb
cbeer: Extracted to #326

PR #315, opened 2019-07-29
Remove explicit color styling in favor of global bootstrap variable

### LINKED ISSUE #312: Blue button gets blue text on keyboard focus
Screen Shot 2019-07-29 at 11 23 25 AM

`a.btn` should retain white text in hover/focus/active states. The background colour is changing appropriately.

mejackreed:
fixes #312


mejackreed:
Happy to pair up to review this to confirm we have the correct behavior everywhere

PR #318, opened 2019-07-29
Display appropriate values for various pieces of patron data for expired users

### LINKED ISSUE #95: Add "Expired" patron standing and implement "no access" labels
@jvine @saseestone What should a user with expired privileges see when they log into the app? Should they be denied access or do they login, see a summary page with a specific status?

jkeck:
Closes #95

Expired User

### DIFF
--- a/app/views/patron/_patron.html.erb
+++ b/app/views/patron/_patron.html.erb
@@ -22,6 +22,8 @@
     <dt class="col-4 col-md-2 font-weight-normal">Privileges expired:</dt>
     <dd class="col-8 col-md-10 expired-date"><%= l(patron.expired_date, format: :short) %></dd>
   <% end %>
-  <dt class="col-4 col-md-2 font-weight-normal">Email:</dt>
-  <dd class="col-8 col-md-10 email"><%= patron.email %></dd>
+  <% if patron.email %>
+    <dt class="col-4 col-md-2 font-weight-normal">Email:</dt>
+    <dd class="col-8 col-md-10 email"><%= patron.email %></dd>
+  <% end %>
 </dl>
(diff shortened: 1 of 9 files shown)


app/views/patron/_patron.html.erb
```diff
@@ -6,18 +6,30 @@
...
+    <dd class="col-8 col-md-10 expired-date"><%= l(patron.expired_date, format: :short) %></dd>
+  <% end %>
+  <% if patron.expired_date && patron.expired? %>
+    <dt class="col-4 col-md-2 font-weight-normal">Privileges expired:</dt>
+    <dd class="col-8 col-md-10 expired-date"><%= l(patron.expired_date, format: :short) %></dd>
+  <% end %>
   <% if patron.borrow_limit %>
     <dt class="col-4 col-md-2 font-weight-normal">Borrower limit:</dt>
     <dd class="col-8 col-md-10 patron-status"><%= pluralize(patron.borrow_limit, 'checkout') %></dd>
+  <% elsif patron.expired? %>
```
jkeck: Some of this `if`/`elsif` logic is getting a bit complicated in this view.  I'm thinking about breaking this out into separate partials and having `to_partial_path` return the appropriate thing based on if the patron is expired, or a fee borrower/proxy user (which I think are the two main distinctions we need to make).

cbeer: I'm good with splitting out a special "expired user" partial. It's the condition on L13-16 that really bugs me.

jkeck: I ended up splitting out a fee_borrower partial in addition to the expired user partial to keep things a bit cleaner.  Now there are just simple `if` statements in there (and I threw in tests for all partials for good measure).

PR #319, opened 2019-07-29
Cleaner views for users without fines, requests, or checkouts

### LINKED ISSUE #267: Add placeholder content if there are no checkouts / requests / fines

tallenaz:
Fixes #267 
  Paired with @camillevilla

### DIFF
--- a/app/views/checkouts/index.html.erb
+++ b/app/views/checkouts/index.html.erb
@@ -6,28 +6,8 @@
 <% end %>
 <% if @checkouts.any? %>
   <% recalled, other_checkouts = @checkouts.partition(&:recalled?) %>
-  <div class="page-section">
-    <h2>Recalled: <%= recalled.length %></h2>
-      <% if recalled.any? %>
-        <div class="d-none d-md-flex row font-weight-bold list-header">
-          <div class="row col-md-4">
-            <div class="col-md-6">Status</div>
-            <div class="col-md-6">Due date</div>
-          </div>
-          <div class="row col-md-8">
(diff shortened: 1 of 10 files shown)


app/views/checkouts/index.html.erb
```diff
@@ -4,72 +4,74 @@
...
+      <%= render other_checkouts %>
+    </ul>
   </div>
-  <ul class="list checkouts list-group">
-    <%= render other_checkouts %>
-  </ul>
-</div>
-
+<% else %>
+  <h2>Checked out: <%= @checkouts.length %><%= " (#{patron_or_group.remaining_checkouts} remaining)" if patron_or_group.remaining_checkouts %></h2>
```
cbeer: I think we still need the `div.page-section` here in order to have appropriate spacing between this and the policies. I see you took a slightly different approach to this in fines and requests, so maybe it's worth just doing the same thing here (putting the `if` condition around the sort + data instead of the whole thing)?

camillevilla: We threw in the a `div.page-section` into the `else` branch. The approach is slightly different here because we don't display `Recalled: 0` (whether they have checkouts or not) if a user has no recalled items.

camillevilla:
@cbeer see comment above. We also split out some partials because the checkouts index page was getting unwieldy.

app/views/checkouts/index.html.erb
cbeer: 👍 WFM.

PR #332, opened 2019-07-30
Update flash message for bulk renewals

### LINKED ISSUE #325: Flash messages for bulk renewals
A misunderstanding on the flash message situation for bulk renewals:  since it's possible for one or more renewals to fail (though presumably not for eligibility reasons), a single summary flash message wasn't feasible. I expected splitting the messages to give a max of 2: 
- "Success! 23 items were renewed"
- "Sorry! 4 items were not renewed"

In fact, each item now gives an individual flash message. This is fine for small 2-3 items, but is overwhelming when there are more than a few; and will quickly obscure the full page. 

At the same time, there's value in the specific information about which items were affected. 

Wondering if we can instead:
- give a single flash message for the successful renewals, and a single message for failures
- list the affected titles in the single message

This would still be a large message for a mass renewal, but it's _less_ large, and a single dismiss action; and the list is more easily scannable than the separate messages.

**Before** (pretending there's a failure which I haven't actually seen)

Screen Shot 2019-07-29 at 1 59 19 PM

**After**

Screen Shot 2019-07-29 at 2 07 15 PM

(Not certain about the text of the failure message - I want something generic. Suggestions?)

jkeck:
Closes #325 

Screen Shot 2019-07-29 at 4 59 01 PM

### DIFF
--- a/app/services/symphony_client.rb
+++ b/app/services/symphony_client.rb
@@ -97,14 +97,14 @@ class SymphonyClient
   end
 
   def renew_items(checkouts)
-    checkouts.each_with_object(success: [], error: []) do |checkout, flash|
+    checkouts.each_with_object(success: [], error: []) do |checkout, status|
       response = renew_item(checkout.resource, checkout.item_key)
 
       case response.status
       when 200
-        flash[:success] << I18n.t('mylibrary.renew_item.success_html', title: checkout.title)
+        status[:success] << checkout
       else
-        flash[:error] << I18n.t('mylibrary.renew_item.error_html', title: checkout.title)
+        status[:error] << checkout
       end
     end
   end
--- a/spec/services/symphony_client_spec.rb
+++ b/spec/services/symphony_client_spec.rb
@@ -120,11 +120,11 @@ RSpec.describe SymphonyClient do
       ]
     end
 
-    it 'returns success + error values for individual renewal requests in symphony' do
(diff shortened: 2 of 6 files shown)


spec/services/symphony_client_spec.rb
```diff
@@ -120,11 +120,11 @@
       ]
     end
 
-    it 'returns success + error values for individual renewal requests in symphony' do
+    it 'returns successfull + errorored titles for individual renewal requests in symphony' do
```
cbeer:
```suggestion
    it 'returns successful + errored titles for individual renewal requests in symphony' do
```

app/services/symphony_client.rb
```diff
@@ -97,14 +97,14 @@ def renew_item(resource, item_key)
...
 
   def renew_items(checkouts)
-    checkouts.each_with_object(success: [], error: []) do |checkout, flash|
+    checkouts.each_with_object(success: [], error: []) do |checkout, status|
       response = renew_item(checkout.resource, checkout.item_key)
 
       case response.status
       when 200
-        flash[:success] << I18n.t('mylibrary.renew_item.success_html', title: checkout.title)
+        status[:success] << checkout.title
```
cbeer: I'm 👍 with what's here, although I wonder about returning the whole checkout and letting the thing that receives it pluck out the title?

jkeck: 👍 that makes sense.

Issue #220, opened 2019-07-23
Set the click areas + hover state for table expand/collapse behavior


cbeer:
- [x] primary row (status, date, title, author, call number) is clickable to expand/collapse
- [x] expanded row is NOT clickable to collapse
- [x] hovered row has a background color of $light-sandstone: #F9F6EF
         - Overdues: default #F9DCB8 - 5% = #FFE9C5
         - Recalls: default #f5c6cb -12% = #FFE5EA
- [x] expanded row has a background color of $light-sandstone: #F9F6EF as long as it is open
         - Overdues: default #F9DCB8 - 5% = #FFE9C5
         - Recalls: default #f5c6cb -12% = #FFE5EA
- [x] buttons/links in the expanded row behave as expected. (!)

cbeer:
Should we make more of the row/list item thing clickable, `.stretched-link` may be our friend.

jvine:
`.stretched-link` is super cool, but in our layout it makes the entire expanded content part of the expand/collapse action, overriding the actions and links that are present in that content. If there's a way to resolve that...

Goal is:
- primary row (status, date, title, author, call number) is clickable to expand/collapse
- expanded row is NOT clickable to collapse
- hovered row has a background color of $light-sandstone: #F9F6EF
- expanded row has a background color of $light-sandstone: #F9F6EF as long as it is open
- buttons/links in the expanded row behave as expected. (!)

![expando](https://user-images.githubusercontent.com/6945691/62093694-25dda300-b22f-11e9-9baa-88a6afbfbd10.gif)

cbeer:
Does that hover + expanded background color apply for all status?

jvine:
Thanks for checking; the sandstone is a jarring change for recalls in particular. Can we use a lightened version of the current background colour?

Overdues:  default  `#F9DCB8`  - 5% = `#FFE9C5`
Recalls:   default  `#f5c6cb` -12% = `#FFE5EA`

(Turns out this fixes a contrast error I hadn't caught before, so.)

jkeck:
Closed by #376

PR #334, opened 2019-07-30
Make item rows expandable by clicking anywhere on the row

cbeer:
Part of #220


cbeer:
In #220, @jvine explicitly said:
> expanded row is NOT clickable to collapse

mejackreed:
👍 thanks I missed that. Just something I felt myself wanting to do :)

jvine:
^ I understand, me too. But I'm concerned about maintaining distinct click targets for the actions in the expando area, to avoid accidentally closing the row when you're trying to renew.

PR #338, opened 2019-07-30
Implement breakpoint truncation changes fixes #330

### LINKED ISSUE #330: Title truncation broken; related column layout change
The title truncation is not behaving quite correctly - it seems to be affected by the height of the call number, which can't really be truncated (reason # 27 why I don't want it). 

Screen Shot 2019-07-29 at 3 42 58 PM

Looking into the truncation issue led me to consider another change, that will make the author & call number more readable in mid-width displays.

- [x] Proposal is to change `col-md-6` to `col-md-12 col-lg-6` for both call number and author. This will create an in-between state where everything else is still in column layout, but the author and call number are stacked. Looks like this:

**Large**

Screen Shot 2019-07-29 at 6 02 06 PM

**Medium**

Screen Shot 2019-07-29 at 6 03 03 PM

**Which seems better than this, no?**

Screen Shot 2019-07-29 at 6 02 53 PM

- [x] Also, given that the call number frequently wraps to 3 or more lines (grr) and that the combination of text size and flex layout on mobile makes the title sometimes quite short, can we change the title truncation from 2 lines to 3? 

- [ ] The strange overflow issue needs to be resolved. It appears that clamp is expecting a line-height of 1.6, and the element being clamped has a line-height of 1.4 - that seems to account for the partially visible lines. But the additional content appearing after the ellipsis is a mystery.

mejackreed:
Fixes #330
The main changes here:
 - Only do the renewable-indicator negative margin when on lg and above screens
 - Move everything from 2 -> 3 line truncation
 - Modify how checkouts stack in md and below sizes

I paired with @jvine on the final details.

Screen Shot 2019-07-30 at 10 52 01 AM
Screen Shot 2019-07-30 at 10 51 47 AM

### DIFF
--- a/app/views/checkouts/index.html.erb
+++ b/app/views/checkouts/index.html.erb
@@ -34,14 +34,14 @@
 
     <div class="d-none d-md-flex row font-weight-bold list-header">
       <div class="row col-md-4">
-        <div class="col-md-6" data-sort="status">Status</div>
-        <div class="col-md-6 active" data-sort="due_date">Due date</div>
+        <div class="col-md-12 col-lg-6" data-sort="status">Status</div>
+        <div class="col-md-12 col-lg-6 active" data-sort="due_date">Due date</div>
       </div>
       <div class="row col-md-8">
         <div class="col-md-7" data-sort="title">Title</div>
         <div class="row col-md-5">
-          <div class="col-md-6" data-sort="author">Author</div>
-          <div class="col-md-6 call_number" data-sort="call_number">Call number</div>
+          <div class="col-md-12 col-lg-6" data-sort="author">Author</div>
+          <div class="col-md-12 col-lg-6 call_number" data-sort="call_number">Call number</div>
         </div>
       </div>
     </div>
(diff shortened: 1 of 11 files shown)


app/views/checkouts/index.html.erb
```diff
@@ -34,14 +34,14 @@
...
       <div class="row col-md-4">
-        <div class="col-md-6" data-sort="status">Status</div>
-        <div class="col-md-6 active" data-sort="due_date">Due date</div>
+        <div class="col-md-12 col-lg-6" data-sort="status">Status</div>
+        <div class="col-md-12 col-lg-6 active" data-sort="due_date">Due date</div>
       </div>
       <div class="row col-md-8">
         <div class="col-md-7" data-sort="title">Title</div>
         <div class="row col-md-5">
-          <div class="col-md-6" data-sort="author">Author</div>
```
cbeer: Do we want to do this same collapse behavior for requests + fines?

mejackreed: Great question. @jvine Do you want the same behavior there?

mejackreed: I've popped on a commit that does it in case we do.

jvine: Yes, we do. (Royal we.)

mejackreed: Great.. I think I've made the changes needed here now

PR #341, opened 2019-07-30
Fix bug where fine payment incorrectly shows Paid

### LINKED ISSUE #340: Fix bug where fine payment incorrectly shows 'Paid'
If there is no `paymentTypeDescription` the default return text value is 'Paid' even if the fee was `FORGIVEN`. We need to check the `paymentTypeID` as well.

jgreben:
Fixes #340 and #347

### DIFF
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -4,26 +4,23 @@
 class Payment
   attr_reader :record
 
-  UNPAID_TYPES = {
-    'AUTOREFUND' => 'Removed',
-    'CANCEL' => 'Cancelled',
-    'CLMS-RTND' => 'Returned',
+  PAYMENT_TYPES = {
+    'AUTOREFUND' => 'Returned',
+    'CANCEL' => 'Removed',
+    'CARDSWIPE' => 'Paid by credit card',
+    'CASH' => 'Paid by cash',
+    'CHECK' => 'Paid by check',
+    'CLMS-RTND' => 'Processing claim',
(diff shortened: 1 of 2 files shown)


app/models/payment.rb
```diff
@@ -42,6 +42,7 @@ def payment_date
 
   def resolution
     return fee_pay_info['paymentTypeDescription'] if fee_pay_info['paymentTypeDescription']
+    return fee_pay_info['paymentTypeID'] if fee_pay_info['paymentTypeID']
```
shelleydoljack: I'm assuming that we always have a paymentTypeID, so do we do anything with it (translate?)? And does the final condition ever get called (return Paid)?

jgreben: We refactored the payment model.

jgreben:
Screen Shot 2019-07-30 at 1 07 53 PM

spec/models/payment_spec.rb
```diff
@@ -58,12 +56,55 @@
...
+  context 'when payment resolution is a PAID_TYPE' do
     it 'has a resolution desctiption' do
-      expect(payment.resolution).to eq 'Payment using credit or debit card via MyAccount'
+      expect(payment.resolution).to eq 'Paid by credit card'
     end
+  end
 
-    it 'can tell if they paid their bill using a card' do
-      expect(payment).to be_paid_fee
+  context 'when no payment resolution is an UNPAID_TYPE' do
```
shelleydoljack: We should update the context to payment type.

PR #356, opened 2019-07-30
Sort the fines in reverse date order (newest first).

### LINKED ISSUE #345: Reverse-sort the payments history
Most recent payments should be at the top of the list (per @jvine)

jkeck:
Closes #345

### DIFF
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -40,6 +40,12 @@ class Payment
     fee_pay_info && Time.strptime(fee_pay_info['paymentDate'], '%Y-%m-%d')
   end
 
+  def sort_key
+    return Time.zone.parse(fee_pay_info['paymentDate']) if fee_pay_info && fee_pay_info['paymentDate']
+
+    Request::END_OF_DAYS
+  end
+
   def resolution
     return fee_pay_info['paymentTypeDescription'] if fee_pay_info['paymentTypeDescription']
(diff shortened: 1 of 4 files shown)


app/models/payment.rb
```diff
@@ -40,6 +40,12 @@ def payment_date
     fee_pay_info && Time.strptime(fee_pay_info['paymentDate'], '%Y-%m-%d')
   end
 
+  def sort_key
+    return Time.zone.parse(fee_pay_info['paymentDate']) if fee_pay_info && fee_pay_info['paymentDate']
+
+    Request::END_OF_DAYS
```
jkeck: Should the payments in history w/o a payment date sort at the top or the bottom?

cbeer: We've sorted missing data to the top everywhere else.

camillevilla: @jkeck can you clarify and provide an example user?

jkeck: I don't have an example handy, but my question/concern comes from how some of the accessors in `Payment` are written.

https://github.com/sul-dlss/mylibrary/blob/ddd8205559a6c61733c973543ba318971d32378b/app/models/payment.rb#L39-L41

Since this method could return `nil` and sorting will compare a `Time` vs. `nil` (which would throw an error) so I'm guarding against that.

PR #368, opened 2019-07-31
Add a sort_key method to the BorrowDirectRequests::Request class so …

jkeck:
…it can be sorted properly w/ other requests.

### DIFF
--- a/app/models/borrow_direct_requests.rb
+++ b/app/models/borrow_direct_requests.rb
@@ -39,6 +39,17 @@ class BorrowDirectRequests
       request_number
     end
 
+    def sort_key(sort)
+      case sort
+      when :title
+        title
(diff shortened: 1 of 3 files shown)


app/models/borrow_direct_requests.rb
```diff
@@ -39,6 +39,17 @@ def key
       request_number
     end
 
+    def sort_key(sort)
+      case sort
+      when :title
+        title
+      when :date
+        ::Request::END_OF_DAYS.strftime('%FT%T')
```
cbeer:
```suggestion
        [::Request::END_OF_DAYS.strftime('%FT%T'), title].join('---')
```

So we fall back to title sort (since all the dates will be the same)

PR #385, opened 2019-07-31
Add a disabled styles to all disabled w/ the disabled attribute

### LINKED ISSUE #360: Custom disabled button styling
Disabled blue button:
- change its background color to **$cool-grey 80%** `rgba(77,79,83,0.8)` 

Screen Shot 2019-07-30 at 1 13 07 PM
Screen Shot 2019-07-30 at 1 13 01 PM

jkeck:
Closes #360 

 

Screen Shot 2019-07-31 at 4 21 08 PM

### DIFF
--- a/app/assets/stylesheets/base.scss
+++ b/app/assets/stylesheets/base.scss
@@ -15,6 +15,15 @@ a {
   }
 }
 
+:disabled {
+  // scss-lint:disable ImportantRule
+  background-color: $sul-btn-disabled-color !important;
+  color: $white;
+  cursor: not-allowed;
+  text-decoration: none;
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/base.scss
```diff
@@ -15,6 +15,15 @@ a {
   }
 }
 
+:disabled {
+  // scss-lint:disable ImportantRule
+  background-color: $sul-btn-disabled-color !important;
```
jkeck: There is an `a.disabled` styling above, but I don't see anywhere we have links w/ a disabled class on it (and I think its intention is different than the disabled button (e.g color is disabled instead of background).

PR #394, opened 2019-08-01
Consider SUL sourced checkouts to be from BorrowDirect…

### LINKED ISSUE #389: In Checkouts, translate source = SUL and remove the SW link
In the Checkouts tab, we need to translate the Source library for "SUL" to "BorrowDirect"

jkeck:
…(and do not attempt to link to the SearchWorks record)

Closes #389

Screen Shot 2019-08-01 at 9 37 11 AM

### DIFF
--- a/app/models/checkout.rb
+++ b/app/models/checkout.rb
@@ -4,6 +4,7 @@
 class Checkout
   attr_reader :record
 
+  BORROW_DIRECT_CODE = 'BORROW_DIRECT'
   SHORT_TERM_LOAN_PERIODS = %w[HOURLY].freeze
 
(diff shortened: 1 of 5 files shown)


app/models/checkout.rb
```diff
@@ -101,7 +102,10 @@ def days_remaining
   end
 
   def library
-    fields['library']['key']
+    code = fields['library']['key']
+    return BORROW_DIRECT_CODE if code == 'SUL'
```
jkeck: This only applies to checkouts.  When an ON_ORDER item is requested its library is SUL so we don't want this translation to be universal.

Issue #402, opened 2019-08-01
Payment processing UX issues, part 2: sponsors


jvine:
When a proxy sponsor makes a Self payment, the issues from #399 apply. 

When a sponsor makes a payment from the Proxies tab, it gets more complicated.

- app returns to the Self tab
- nav status shows "payment processing"
- Proxies tab still shows the full payable amount

<img width="400" alt="Screen Shot 2019-08-01 at 1 44 36 PM" src="https://user-images.githubusercontent.com/6945691/62329047-9a038b00-b469-11e9-8d17-e7f382dc5ed7.png">

HOWEVER!!! Switching to the proxies tab, the bills are still present, and ~1 hour later workflows shows the bills were not paid. So that $340 payment went... somewhere.

<img width="400" alt="Screen Shot 2019-08-01 at 1 45 17 PM" src="https://user-images.githubusercontent.com/6945691/62329069-ac7dc480-b469-11e9-973c-4e213b940850.png">

camillevilla:
Pairing session with @jgreben @dlrueda reveals: 
We currently send the sponsor's libraryid with the API request. However, proxies fines are only associated with the proxy's libraryid. When a faculty member tries to pay a proxy's fine, we need to ensure the proxy's libraryid is sent with the API request instead.

camillevilla:
Results of discussion: 
Faculty member should NOT see a pay button on the `Proxies` tab. 
Faculty member should still see a pay button on the `Self` tab.

camillevilla:
@jvine and @saseestone will handle placeholder text to use in the mean time.

jvine:
Placeholder text:

> Fines can be paid in My Library Account only by the borrower who accrued them. 
To pay group fines using iJournal, <popup>contact Circulation & Privileges</popup>.

<img width="720" alt="Screen Shot 2019-08-01 at 3 55 38 PM" src="https://user-images.githubusercontent.com/6945691/62332528-eb654780-b474-11e9-9089-3e46ca965d18.png">

camillevilla:
Further investigation in Symphony workflows revealed that sponsors are currently unable to pay proxies' fines. 

Interrim fix has been implemented via #405 and brings us to parity on this feature.

To achieve desired behavior, further work in Symphony cgi-script needs to be done. We should retain this ticket to understand the desired UX, but I'm moving this out of the ready column in the meantime.

PR #405, opened 2019-08-01
Group fine payments

jgreben:
Part of #402 

A temporary solution to the fact that sponsors cannot pay for their proxys' fines. Further work in Symphony cgi-script is required to allow these types of payments.


jgreben:
Screen Shot 2019-08-01 at 4 09 09 PM

PR #409, opened 2019-08-02
Does not render the fines table when viewing other proxies fines

### LINKED ISSUE #388: Proxy accounts should see and pay only the fines associated with their account
Due to privacy concerns, proxy users should only be able to view and pay the fines associated with their proxy account.  They should not be able to view or pay the fines associated with other proxy accounts. 

Sponsors should be able to view and pay fines for both their individual account and all the proxy accounts. (#402)

camillevilla:
Closes #388

## Before
Proxy23 can see Proxy21's fines.
Screen Shot 2019-08-01 at 5 36 28 PM

## After
Proxy23 cannot see Proxy21's fines, does not have a pay button, and gets a notification.
Screen Shot 2019-08-01 at 5 36 12 PM

### DIFF
--- a/app/views/fines/index.html.erb
+++ b/app/views/fines/index.html.erb
@@ -9,11 +9,7 @@
   <h2>Payable: <%= number_to_currency(@fines.sum(&:owed)) %></h2>
 
   <% if @fines.any? && !params[:payment_pending] %>
-      <% unless params[:group] && patron.sponsor? %>
-        <div class="mb-3">
-          <%= render 'fines/pay_all_button' %>
-        </div>
-      <% else %>
+      <% if params[:group] && patron.sponsor? # temporary fix until Symphony CGI scripts are updated %>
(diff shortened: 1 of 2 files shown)


app/views/fines/index.html.erb
```diff
@@ -9,33 +9,40 @@
...
+        <div class="mb-3">
+          <%= render 'fines/pay_all_button' %>
         </div>
-      </div>
-      <ul class="fines list-group">
-        <%= render @fines %>
-      </ul>
+      <% end %>
+
+      <% unless params[:group] && !patron.sponsor? %>
```
cbeer:
```suggestion
      <% if !params[:group] || patron.sponsor? %>
```

This might be  a little clearer? Or, maybe:

```
      <% if patron_or_group.is_a?(Patron) || patron.sponsor? %>
```

Issue #155, opened 2019-07-17
Fines: Translate "Bill reasons" to nice labels


camillevilla:
Split out from #86

Helpful info from @saseestone:
---
So the list that Chris added above must be the webservices labels for the "Bill Reasons" that I see in the policies list in WorkFlows.  (slightly different wording...so we'd want @shelleydoljack to confirm that we're not using a WorkFlows policy...differently...than web services is expecting.  (e.g. our odd use of "group ID" in policies vs. the auth id field in a user record.)

Would really appreciate it if @jvine can review my suggested translations before they are put in the app. :)

We use the following "Bill Reasons":
Symphony policy "Bill Reason" name -> Web services label (?) -> Sarah's recommended translation
OVERDUE -> ITEM_OVERDUE -> Overdue item
RECALLOVD -> RECALLED_ITEM_OVERDUE -> Overdue recall
RESERVEOVD -> RESERVE_ITEM_OVERDUE -> Overdue course reserve
LOST -> LOST_ITEM -> Lost item
REPLCMENT -> ? -> Lost item
LOST-ILL -> ? -> Lost interlibrary loan item
CLAIM-LOST -> ? -> Lost item
CLAIM-FEE -> ? -> Processing fee
PROCESSFEE -> PROCESSING_FEE -> Processing fee
PROCESSING -> ? -> Processing fee
DAMAGED -> ? -> Damaged item
PRIVILEGES -> PRIVILEGE_FEE -> Privileges fee
LOSTCARD -> ? -> Lost card fee
LOSTKEY -> ? -> Lost key fee

Older bills that we have in the database and *could* should up in My Account:
BILLED-OD -> ? -> Lost item
BILLING-FEE -> ? -> Processing fee
PRE-NOTIS -> ? -> Lost item
PRE-UNICRN -> ? -> Lost item
RECAL-BILL -> ? -> Lost item
RECAL-FEE -> ? -> Overdue recall
RECOD-MBIL -> ? -> Lost item
RECOD-MFEE -> ? -> Processing fee
RECOD-MOD -> ? -> Lost item

Delivered policies that exist in Symphony but we don't use at Stanford...but *could* should up sometime/somewhere:
BADCHECK -> ? -> Privileges fee
LOSTRECALL -> ? -> Lost item
MISC -> ? -> Overdue item

PR #411, opened 2019-08-02
Render correct bill reasons

### LINKED ISSUE #406: Bill reasons in payment history don't match current bills
See #155 for list

"Overdue item" (correct) vs. "Overdue" (not in the translation list)

Screen Shot 2019-08-01 at 4 06 59 PM

"Damaged item" (correct) vs. "Damage fine" (not in the translation list)

Screen Shot 2019-08-01 at 4 08 59 PM

tallenaz:
Fixes #406

### DIFF
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -23,6 +23,23 @@ class Payment
     'XREMBILL' => 'Removed'
   }.freeze
 
+  BILL_REASONS = {
+    'OVERDUE' => 'Overdue item',
+    'RECALLOVD' => 'Overdue recall',
+    'RESERVEOVD' => 'Overdue course reserve',
+    'LOST' => 'Lost item',
+    'REPLCMENT' => 'Lost item',
+    'LOST-ILL' => 'Lost interlibrary loan item',
+    'CLAIM-LOST' => 'Lost item',
+    'CLAIM-FEE' => 'Processing fee',
+    'PROCESSFEE' => 'Processing fee',
+    'PROCESSING' => 'Processing fee',
+    'DAMAGED' => 'Damaged item',
+    'PRIVILEGES' => 'Privileges fee',
+    'LOSTCARD' => 'Lost card fee',
+    'LOSTKEY' => 'Lost key fee'
+  }.freeze
+
   def initialize(record)
     @record = record
   end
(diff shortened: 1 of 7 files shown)


tallenaz:
For user `super1`,
Before:
Screen Shot 2019-08-01 at 7 15 14 PM
After:
Screen Shot 2019-08-01 at 7 12 24 PM

app/models/payment.rb
```diff
@@ -32,7 +32,14 @@ def key
   end
 
   def bill_description
-    record['billReasonDescription']
+    case record['billReasonDescription']
```
mejackreed: We do some other type of translations using a hash. Is it possible to keep a similar pattern for that?

Translation:
https://github.com/sul-dlss/mylibrary/blob/master/config/application.rb#L41-L69

Used:
https://github.com/sul-dlss/mylibrary/blob/a4b72fe7221fe57331733a283f8f6dddb799cfee/app/helpers/application_helper.rb#L63

jvine:
It looks like only these two translations were added? These were just 2 examples I happened to find, but there were several other changes we made in the translations. All the Paid reason translations should match the Payable reasons.

app/models/payment.rb
tallenaz: Thanks for catching that. Done in https://github.com/sul-dlss/mylibrary/pull/411/commits/d3391b7630a140b8dea230e87ec3a11853688f68

app/models/payment.rb
```diff
@@ -32,7 +32,7 @@ def key
   end
 
   def bill_description
-    record['billReasonDescription']
+    Mylibrary::Application.config.bill_reasons[record['billReasonDescription']] || record['billReasonDescription']
```
mejackreed: Can we also move this to a helper, so other instances could access this native value also?

tallenaz:
@jvine @saseestone the previous translation map was translating (we think) the fine status. The `bilReasonDescription` we're using in the "Reason" column from Symphony seems different from the fine status values. Is there a pre-existing mapping we can use for `billReasonDescription`?

saseestone:
@tallenaz  it looks to me like in the fines tab, for Payable bills, we're using translations for "Reason" as outlined in #155.

But for Payment History (hopefully soon to become labelled just "History"), we're using the bilReasonDescription from the Symphony policy list for "Bill Reason".   Does that jive with what you're seeing in the code?

We should not be using the Symphony policy description in either case.  We should be using the translations as outlined in #155 .  Doable?

tallenaz:
@saseestone this now uses a mapping from #155 instead of `billReasonDescription`.

PR #413, opened 2019-08-02
Extract various constant/config maps into settings 

cbeer:
🤷‍♂


mejackreed:
My mental model of these things are usually..

Settings == things that private or that we want to change host/environment specific
config == semi-static things and translation maps

IDK if others also have this, but i would be 👎 to 👌 with this change as it changed some of that up.

camillevilla:
Feeling neutral about this, but next steps should incorporate the `BILL_REASONS` addded in #411

cbeer:
ok. let's poll the team next week. i'm happy to just close this if no one else is as enthused as i am.

mejackreed:
Just adding this note here.. We also need to be cognizant of how config gem handles merges by default. Might just not be the expected understanding: https://github.com/railsconfig/config#merge-customization

PR #429, opened 2019-08-02
Add an okcomputer check to ping the symphony oracle database

### LINKED ISSUE #416: Add an okcomputer check for connecting to oracle

cbeer:
Fixes #416

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -11,4 +11,14 @@ class SymphonyClientCheck < OkComputer::Check
   end
 end
 
+# OKComputer check that checks if we have a connection to the symphony oracle db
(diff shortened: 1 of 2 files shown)


config/initializers/okcomputer.rb
```diff
@@ -11,4 +11,14 @@ def check
   end
 end
 
+# OKComputer check that checks if we have a connection to symws
```
shelleydoljack: This checks the connection to Symphony database server, not web services. How often does this check occur? When app is deployed?

cbeer: Nagios checks this endpoint (every 5 minutes?) as part of its normal health checks.

shelleydoljack: Thanks!

PR #433, opened 2019-08-02
Implements new UX modifications for payment workflow

### LINKED ISSUE #399: Payment processing UX issues, part 1: individuals
For an individual user, the state of the UI on returning from a successful fine payment is really confusing.

The return page shows:
- nav status:  "**payment processing**" (...okay, but see comments)
- flash "Success! $41.00 paid..." (correct)
- Payable: **$41.00**  (???  contradicts nav status)
- no payment button (correct, but confusing with Payable amount shown above)
- fines are listed in payment history because the response was quick; I know this won't always be the case.

Screen Shot 2019-08-01 at 1 57 10 PM

When I refresh the page, it shows:
- nav status:  "**payment processing**"  (now it's wrong)
- no flash  (correct)
- Payable: **$0.00**  (correct)
- no payment button (correct)
- fines are listed in payment history (correct)

Screen Shot 2019-08-01 at 1 58 12 PM

Both of these states are inaccurate and confusing.

1. The "payment processing" text in nav status is too long and breaks on mobile [Edit: I thought "processing" would work, but it's still too long. "pending" fits.]
  
2. The nav status and Payable headers should always match - both should be "pending" when the payment has not been processed, and both should be "$0.00" when it has. 

3. If there is no content in the Payable fines list - i.e. the fines are already showing in Payment history - can the nav status and Payable header both show $0.00 right away? And show "pending" only when the fines are actually still listed?

mejackreed:
Fixes #399 

finesux


mejackreed:
I also think this may `fix` #402. But haven't tested it yet.

app/helpers/payments_helper.rb
```diff
@@ -2,19 +2,4 @@
 
 # Helper module for Fine Payments
 module PaymentsHelper
```
cbeer: Should we just remove the helper + spec file entirely?

mejackreed: Perhaps.. I left it in incase others were rebasing and there were problems. I'll remove it now.

app/models/patron.rb
```diff
@@ -137,7 +138,19 @@ def checkouts
   end
 
   def fines
```
cbeer: It seems like a patron unit test is probably in order?

app/helpers/payments_helper.rb
mejackreed: Pushed up a commit that removes this.

app/models/patron.rb
mejackreed: Good call. I've pushed this up.

cbeer:
👏 👏 👏

PR #437, opened 2019-08-03
Requests for on-order items do not have queue length or position.

shelleydoljack:
Some holds, like title-level for things on-order return nil values for queueLength and queuePosition:
Screen Shot 2019-08-02 at 10 03 39 PM

This makes the waitlist position "Unknown" for these situations:
Screen Shot 2019-08-02 at 10 33 04 PM

### DIFF
--- a/app/models/request.rb
+++ b/app/models/request.rb
@@ -77,6 +77,8 @@ class Request
   end
 
   def waitlist_position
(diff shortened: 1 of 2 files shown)


cbeer:
This seems like an improvement over broken text; @jvine, can you weigh in on the language (or if we should have some alternate text here instead?)

PR #443, opened 2019-08-05
Stop loading spinner if no payment history is available

### LINKED ISSUE #436: Consider hiding fine history for users without such history
In the fines tab, if a user has no history of fines, when the "Show history" button under "History" is clicked, the button will show a spinning "Loading" indicator and hang in that state.

camillevilla:
Fixes #436 

If a user has no fines, stop the loading spinner and returns the message:

Screen Shot 2019-08-05 at 9 59 25 AM

### DIFF
--- a/app/views/payments/index.html.erb
+++ b/app/views/payments/index.html.erb
@@ -1,8 +1,7 @@
-<% if @payments.any? %>
-  <div id="payments" class="page-section">
-    <div class="d-flex justify-content-between">
-      <h2>History</h2>
-
+<div id="payments" class="page-section">
+  <% if @payments.any? %>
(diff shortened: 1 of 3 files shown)


app/views/payments/index.html.erb
```diff
@@ -1,8 +1,7 @@
-<% if @payments.any? %>
-  <div id="payments" class="page-section">
-    <div class="d-flex justify-content-between">
-      <h2>History</h2>
-
+<div id="payments" class="page-section">
+  <% if @payments.any? %>
+  <div class="d-flex justify-content-between">
```
mejackreed: Should we indent this?

PR #454, opened 2019-08-07
Change the source library for requests to the owning library; fixes #453

### LINKED ISSUE #453: In Requests, we should use the owning library as the "Source" library
Looks like in the Requests tab, we're using the placedLibrary to determine "Source".  The problem is most requests placed in SearchWorks have a placedLibrary = SUL.  (As opposed to the fixture users that I created, which all inherited my placedLIbrary = GREEN)

Instead, we should be using the owning library for "Source".  Possible?

In the example from the snapshot 👇 the first item has owning library = GREEN.  Second item has owning library = SAL3. 

image

cbeer:


cbeer:
I've confirmed with @saseestone that we're pulling the data from the right location.

PR #461, opened 2019-08-09
Add barcode to checkouts...

jkeck:
...(and fines? not sure we'll get barcode for fines, but if it is retuned we'll display it)

barcode


cbeer:
I beat you to it in https://github.com/sul-dlss/mylibrary/pull/459. More or less the same idea, but I added couple tests, so 🤷‍♂

PR #476, opened 2019-09-09
Add a configurable flash message for internal pages

camillevilla:
```
internal_pages_flash_message_html: |-
  <p>As of Monday, 9/9/19, we are having an issue with renewing materials due on 9/23/19.  Renewals will be available on Tuesday, 9/10/19.</p>

internal_pages_flash_message_config:
  - summaries
  - checkouts
```

🎃 squash these please

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,6 +3,7 @@
 # :nodoc:
 class ApplicationController < ActionController::Base
   helper_method :current_user, :current_user?, :patron, :patron_or_group, :symphony_client
+  before_action :set_internal_pages_flash_message
 
   def current_user
     session_data = request.env['warden'].user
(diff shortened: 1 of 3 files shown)


app/controllers/application_controller.rb
```diff
@@ -57,4 +58,14 @@ def logout_user!
   def item_details
     {}
   end
+
+  def set_internal_pages_flash_message
+    return unless Settings.internal_pages_flash_message_html && action_name == 'index'
+
+    Settings.internal_pages_flash_message_config.each do |controller|
```
jkeck: This is all good, but just want to point out that when we want to remove the messages we'll need to set the `internal_pages_flash_message_config` config to an empty array (e.g. not null) since this `.each` assumes an enumerable object.

camillevilla: Good catch. Added an empty array to the spec context where nothing is configured.

camillevilla:
Ok! I rebased this branch to shake off some old commits.

PR #489, opened 2019-10-07
Removes flash message announcing new My Library Account.

shelleydoljack:
@saseestone Let us know when you'd like the flash message to be removed.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -14,12 +14,7 @@ RECAPTCHA:
   SITE_KEY: 6Lc6BAAAAAAAAChqRbQZcn_yyyyyyyyyyyyyyyyy
   SECRET_KEY: 6Lc6BAAAAAAAAKN3DRm6VA_xxxxxxxxxxxxxxxxx
 
-home_page_flash_message_html: |-
-  <h1 class="h3 alert-heading">Welcome to the new My Library Account!</h1>
-
-  <p>My Account has been completely re-designed. It shows more information, highlights items that need attention, requires fewer steps, and is easier to use on mobile devices. <a href="https://library.stanford.edu/blogs/stanford-libraries-blog/2019/08/new-my-library-account">&raquo; Find out more</a></p>
-
-  <p class="mb-0">Use the feedback link to let us know what you think!</p>
+home_page_flash_message_html:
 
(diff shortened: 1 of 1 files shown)


saseestone:
Anytime is good.  We've had it up long enough.

app/controllers/sessions_controller.rb
```diff
@@ -63,7 +63,7 @@ def destroy
   private
 
   def set_home_page_flash_message
-    return unless Settings.home_page_flash_message_html
+    return if Settings.home_page_flash_message_html.blank?
```
jgreben: I think this can be left as it was before, and only the `home_page_flash_message_html` should be removed. This way, in the future if we want to add a different message, we just need to add a new message.

config/settings.yml
```diff
@@ -14,12 +14,7 @@ RECAPTCHA:
   SITE_KEY: 6Lc6BAAAAAAAAChqRbQZcn_yyyyyyyyyyyyyyyyy
   SECRET_KEY: 6Lc6BAAAAAAAAKN3DRm6VA_xxxxxxxxxxxxxxxxx
 
-home_page_flash_message_html: |-
-  <h1 class="h3 alert-heading">Welcome to the new My Library Account!</h1>
-
-  <p>My Account has been completely re-designed. It shows more information, highlights items that need attention, requires fewer steps, and is easier to use on mobile devices. <a href="https://library.stanford.edu/blogs/stanford-libraries-blog/2019/08/new-my-library-account">&raquo; Find out more</a></p>
-
-  <p class="mb-0">Use the feedback link to let us know what you think!</p>
+home_page_flash_message_html:
```
jgreben: I think this should just be:
```
internal_pages_flash_message_html: |-
```

This way only the html beneath would need to be removed or added the next time we need a message.

jgreben: Copy paste typo...sorry!

jgreben: Sorry, this should still be
```
home_page_flash_message_html:
```

app/controllers/sessions_controller.rb
shelleydoljack: I changed it back to how it was.

PR #495, opened 2019-10-23
Handle case where the API returns no requests.

jkeck:
Was trolling through some HB reports and noticed [this error](https://app.honeybadger.io/projects/62116/faults/56114076).  Not sure if this happened during a symphony outage and this was just the first symptom of it, so this may not even be helpful (so feel free to take or leave it).

### DIFF
--- a/app/models/patron.rb
+++ b/app/models/patron.rb
@@ -227,7 +227,7 @@ class Patron
   end
 
   def symphony_requests
-    fields['holdRecordList'].map { |request| Request.new(request) }
+    (fields['holdRecordList'] || []).map { |request| Request.new(request) }
   end
 
   def fields


shelleydoljack:
@jkeck I like this idea of handling cases where web services is not returning the field for some reason. I wonder if the patron experience will be okay if the patron logs in and knows they have requests but see 0 requests. Will they submit feedback? Will it be confusing if we don't see a honey badger error that corresponds to the feedback when trying to troubleshoot? @saseestone any thoughts?

jkeck:
¯\\_(ツ)_/¯ I hear ya.  I just saw a 🐛 and 🔪

It seems like we may want to be able to capture for when the API is misbehaving like this and send the user to a systems status page (that would include the fact that the MyLibrary Account system was experiencing an error), but until we can do that, the sort of change in this PR may not be desirable

PR #496, opened 2019-10-25
Add a more descriptive error message for SUNet ID users...

### LINKED ISSUE #490: Failure message possible for users with no library account at login
We've had 3 feedback messages from users that are trying to log into My Library Account and they don't have a Symphony user account.  Can we work on failure message for this particular failure?

Is it a specific failure?

jkeck:
...who fail to authenticate in Symphony.

Plus bonus home page updates!

Closes #490 

Screen Shot 2019-10-25 at 12 08 51 PM

### DIFF
--- a/app/assets/stylesheets/base.scss
+++ b/app/assets/stylesheets/base.scss
@@ -47,6 +47,7 @@ h1 {
   color: $sul-h1-font-color;
   font-weight: 300;
   line-height: 1.2em;
+  margin-bottom: 30px;
 }
 
 h2 {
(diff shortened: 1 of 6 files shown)


app/assets/stylesheets/base.scss
```diff
@@ -47,6 +47,7 @@ h1 {
   color: $sul-h1-font-color;
   font-weight: 300;
   line-height: 1.2em;
+  margin-bottom: 30px;
```
jkeck: @jvine we added this to all `<h1>`s.  Is that okay? (It shows up on the login by pin / reset pin pages as well).

jvine: Yes, nice.

Issue #501, opened 2019-11-22
When patron information is empty, show a helpful error page


jgreben:
Sometimes for various reasons SymWS returns nil for a patron info request. This can happen, for example, when the person has an item in their request or checkouts list that is locked during a Symphony reindexing update during the evening. Here is a message that was captured using verbose logging at the same time a Honeybadger error was reported

`undefined method `dig' for nil:NilClass` (for `fields.dig`)
corresponds to:
```
{"messageList":[{"code":"hatErrorResponse.225","message":"The bibliographic record is locked for text update"}]}
```
We need to redirect to a helpful page explaining what might be going on in for various error codes.

jkeck:
Might be somewhat related to #495

thatbudakguy:
closing since we're no longer using symphony.

PR #506, opened 2019-12-11
Handle various symphony web service response failures

### LINKED ISSUE #504: System maintenance error message needed when WebServices is down
There's been some discussion around what error message displays to the user when WebServices is down or records are locked for update (#501) and (#495).  We (Shelley, Darsi, Sarah) thought the 'down for system maintenance' error would display when LibSys brought webservices down to do a rebuildtext on Nov. 28th.  But instead, we got the 'there's a problem and someone has been alerted.  Try again soon.' 

We have a webservices upgrade planned for Dec. 14th, and potentially another rebuildtext over Winter Break.  We'll need a way to show the 'down for system maintenance' message to users at least during the webservices upgrade.

jgreben:
Fixes #504  

We already handle the scenario where Symphony is down but web services is still running. This is handled using the `symphony_client.ping` method.

There are two additional scenarios we discovered that need the "temporarily unavailable" message to be displayed:

 - When symws is not responsive, whether or not symphony itself is running, such as when Tomcat is down. When this happens, the request for a session token returns some html that describes a 503, and the method that parses the expected JSON for a session token fails.

- ~~When symphony and symws are up and running but symphony is in the middle of performing a rebuildtext process and it locks a patron record because that patron has an item in their checkouts that is being reindexed. In this scenario the patron record is returned as `nil`.~~
(👆 this needs further investigation and should possibly be handled in a subsequent PR.)

We can rescue the JSON parsing errors and return `nil` in the failing symphony_client methods.  We can also check for the ping as a before hook in the application controller so that it will check on subsequent requests for any active controller.


jgreben:
Screen Shot 2019-12-11 at 2 30 45 PM

saseestone:
Josh can you capitalize "My" in the "My Library Account is unavailable..."

jgreben:
@jkeck or @cbeer would you mind taking a look at this PR?

app/services/symphony_client.rb
```diff
@@ -34,10 +34,13 @@ def login_by_sunetid(sunetid)
 
   # get a session token by authenticating to symws
   def session_token
-    @session_token ||= begin
+    @session_token ||=
+      begin
```
jkeck: This indentation seems...odd.  Is rubocop complaining about something here?

app/services/symphony_client.rb
```diff
@@ -34,10 +34,13 @@ def login_by_sunetid(sunetid)
...
       response = request('/user/staff/login', json: Settings.symws.login_params, method: :post)
 
       JSON.parse(response.body)['sessionToken']
+      rescue StandardError
```
jkeck: Can we rescue something more specific than `StandardError` here?  What is the thing that actually fails in this scenario?

app/services/symphony_client.rb
```diff
@@ -65,7 +69,10 @@ def patron_info(patron_key, item_details: {})
     })
 
     JSON.parse(response.body)
+  rescue StandardError
```
jkeck: Can we rescue something more specific than `StandardError` here? What is the thing that actually fails in this scenario?

app/services/symphony_client.rb
jgreben: Yes, rubocop made me do it!

`rescue...is not aligned with begin`

app/services/symphony_client.rb
jgreben: Yes, this can be `JSON::JSONError`

app/services/symphony_client.rb
jgreben: I actually removed this check because after @shelleydoljack and I looked into the problem deeper, we are not sure that this is where to do the check, so we decided to leave this out of this PR and try to address the issue on a different PR to for #501 and #495

jgreben: Yes, this can be `JSON::ParserError`

app/services/symphony_client.rb
```diff
@@ -64,8 +68,13 @@ def patron_info(patron_key, item_details: {})
       ].join(',')
     })
 
-    JSON.parse(response.body)
+    begin
+      JSON.parse(response.body)
+    rescue StandardError
```
jkeck: Should this be a similar `rescue JSON::ParserError` as above?

jgreben: Ideally, but it was breaking the test because it did not expect a body even if provided. I got it to work , however, by removing `body` which do not seem to be necessary.

app/controllers/summaries_controller.rb
```diff
@@ -10,5 +10,9 @@ class SummariesController < ApplicationController
   # GET /summaries.json
   def index
     @patron = patron
+
+    redirect_to unavailable_path if patron.record.nil?
```
jkeck: Is it only the summaries page where we would want to do this redirect?  Could a user theoretically be on one of the other tabs (e.g. Checkouts) when this occurs?

jkeck: Is there a better name for what is meant w/ `patron.record.nil?`?  Is this a state of our patron that we can give a name and call that?  Like, `patron.fubar?` (not really though 😅)

app/services/symphony_client.rb
```diff
@@ -20,7 +20,7 @@ def login(library_id, pin)
       password: pin
     })
 
-    JSON.parse(response.body)
+    JSON.parse(response)
```
jkeck: Huh, this is interesting.  I didn't realize that the JSON or HTTP libraries worked like that.

app/services/symphony_client.rb
```diff
@@ -64,16 +68,21 @@ def patron_info(patron_key, item_details: {})
       ].join(',')
     })
 
-    JSON.parse(response.body)
+    begin
+      JSON.parse(response)
+    rescue JSON::ParserError
+      nil
```
jkeck: I don't feel super strongly one way or the other, but do we want to know when/that this is occurring?  If this were to ever happen w/o us knowing, should we get a notification in Honeybadger?

app/services/symphony_client.rb
jkeck: yuck

app/controllers/summaries_controller.rb
jgreben: I think the Summaries page is (mostly) right. This should happen as soon as they log in, and it's with the `@patron.record` object that we first know something is amiss. You can't get to one of the other tabs directly if you are not already logged in. But yeah... If you are already logged in and then web services goes down and you then refresh on one of the other tabs, the change here would not capture the disconnect and you would see the generic rails error.

app/services/symphony_client.rb
jgreben: I can't see how it does according to the http gem documentation, but it works..

app/services/symphony_client.rb
jgreben: @shelleydoljack also mentioned this. My response to her was that currently we are not (and not really able) doing anything about it if it happens for some unknown SirsiDynix reason. Also when it happens every day during the Symphony restart do we want to get those alerts potentially every day? On the other hand, it might be good to have the data that is captured if someone complains. I can hook it up of you both want.

app/controllers/summaries_controller.rb
jgreben: I made a helper method called `patron_info_unavailable?`. It could also be called from the other tabs if you thing those 👆 scenarios might be worth capturing.

jgreben: I made a helper method called patron_info_unavailable?. It could also be called from the other tabs if you thing those 👆 scenarios might be worth c

jkeck: I'll leave it up to you if you think it's worth doing to the other controllers.  (and if we did that, it might indicate that the unavailable endpoint might make sense as its own controller)

jkeck: And just to be clear wrt a "helper", my comment is more leaning towards some method being on the patron class that you can ask if the patron is problematic.

app/services/symphony_client.rb
jkeck: It makes me feel a little nervous that it was so unexpected.  While I suppose tests would break, it seems like this might be a potential point of confusion (particularly if we change http connection gems or JSON parsers).

I might poke at this for a few mins to see if I can figure out what is going wrong when we use `response.body`.

app/services/symphony_client.rb
jkeck: I pushed up a commit that adds the `.body` back.  I feel like it's a bit more idiomatic and less prone to potential issues down the road.

The test that was failing needed to mimic the `HTTP::Response` interface when stubbing the `authenticated_request` method.

app/services/symphony_client.rb
```diff
@@ -34,10 +34,14 @@ def login_by_sunetid(sunetid)
...
   # get a session token by authenticating to symws
   def session_token
-    @session_token ||= begin
+    @session_token ||=
+      begin
       response = request('/user/staff/login', json: Settings.symws.login_params, method: :post)
 
       JSON.parse(response.body)['sessionToken']
+      rescue JSON::ParserError
+        Honeybadger.notify('Unable to connect to Symphony Web Services.')
```
jgreben: @shelleydoljack @jkeck Do we want to add a Honeybadger config to ignore these notifications? The purpose would  only be to have it logged for reference if we need it for troubleshooting or follow-up.

jkeck: I would leave it up to you all as to if that is something you care about being notified about.  It can always be removed or added later, just wanted to raise the question (as this will no happen completely w/o our knowledge once we are rescuing).

jkeck:
Also, I guess I now have a commit on this PR, so will need to get another set of 👀 on this for review/merging.  

🏃

app/services/symphony_client.rb
jgreben: I guess I am confused about the difference between ignoring a Honeybadger alert by using the configuration, and ignoring an alert (i.e. `is:ignored`) in the Honeybadger interface. I assumed these were the same thing, but they are not(?)

I think we should just go ahead and get the notifications for the time being.

app/controllers/application_controller.rb
```diff
@@ -32,6 +32,12 @@ def patron_or_group
 
   private
 
+  def check_unavailable
+    return if request.path == unavailable_path || request.path == root_path
```
shelleydoljack: Please add a comment to explain what we're doing here.

PR #523, opened 2020-03-13
Update global alert text w/ messaging about request service.

jkeck:
Screen Shot 2020-03-13 at 4 18 35 PM


spec/models/checkout_spec.rb
```diff
@@ -79,6 +79,8 @@
       end
 
       it 'returns the number of days the item is overdue' do
+        # Set "now" to a spcific date and time so we don't this test passes consistently
+        allow(Time.zone).to receive(:now).and_return(Time.zone.parse('2020-03-11').noon)
```
mejackreed: 👏

PR #549, opened 2020-06-04
Add a bit  more granular feature flags for the various schedule once …

jkeck:
…links.

The theory being that because we allow visits to Green doesn't necessarily mean we'll be ready for pickup (even if a valid user has material for pickup) because the library staff may not be ready.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -35,3 +35,9 @@ BORROW_DIRECT_CODE: BORROW_DIRECT
 
 schedule_once:
   enabled: false
+  green_pickup:
+    enabled: false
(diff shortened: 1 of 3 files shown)


config/settings.yml
```diff
@@ -35,3 +35,9 @@ BORROW_DIRECT_CODE: BORROW_DIRECT
 
 schedule_once:
   enabled: false
```
jkeck, line 37: I've kept this top-level flag as a convenience to not have to check both accessors before rendering a wrapping element.  Would be happy to 🔪 this though.

PR #556, opened 2020-06-08
Enables new cops and updates rubocop todo.

shelleydoljack:
We can enable new cops.


.rubocop.yml
```diff
@@ -9,6 +9,14 @@ AllCops:
   TargetRubyVersion: 2.6
   Exclude:
       - vendor/bundle/**/*
+  NewCops: disable
```
jgreben, line 12: Why not enable all new cops so that we get the benefit of keeping the code up to date with the latest style suggestions, and if there is a style that we want to override, then include it here with `Enabled: false`?

shelleydoljack: Hah! I was worried there would be a lot of corrections, but there weren't. I should have checked first.

Issue #557, opened 2020-06-12
Add scheduler for EAL


anarchivist:
> EAL anticipates opening up for visits by appointment on Monday, Jun 22.
Hours would be 10 - 2 (vs 3 pm for Green)
Appointment windows would be 1 hour each, with 12 slots per window (max capacity of 48 people for the day)
Eligible group is the same as Green access

* [x] Create ScheduleOnce event type and booking page (`StanfordLibrariesEastAsiaLibraryEntry`)
* [x] Create shared calendar and provide access to EAL staff who need to see it 
   * [x] SNOW ticket for resource calendar (`RITM00126051`/`REQ00125237`)
   * [x] add calendar to @anarchivist outlook
   * [x] grant access to EAL staff
* [x] Limit appointments to start on 6/22
* [x] Configure My Account to show embedded booking modal

jvine:
My inclination for My Account would be to replace the current "Schedule access to Green Library" button with a drop-down button where we have the flexibility to add more libraries, rather than creating a stack of individual buttons. 

Click button, show list of available libraries, click library name in list, open modal. 

<img width="178" alt="Screen Shot 2020-06-12 at 4 04 27 PM" src="https://user-images.githubusercontent.com/6945691/84680671-e58b2000-aee7-11ea-864a-750b31658222.png">

<img width="187" alt="Screen Shot 2020-06-12 at 4 04 18 PM" src="https://user-images.githubusercontent.com/6945691/84680672-e623b680-aee7-11ea-85c7-204ad5cd9ecb.png">

PR #560, opened 2020-06-16
Add a drop down for scheduling visits when a user has...

jkeck:
... access to multiple libraries

Connected to #557 

### TODO
- [x] Update `routes.rb` with appropriate `oncehub_id` for EAL visits

## One library available/configured
with only one library available/configured

## Multiple libraries available/configured
with multiple libraries available/configured

### DIFF
--- a/app/helpers/summaries_helper.rb
+++ b/app/helpers/summaries_helper.rb
@@ -18,12 +18,29 @@ module SummariesHelper
     )
   end
 
+  def schedule_once_link_or_dropdown
+    return 'Not eligible during current phase of Research Restart Plan' if library_schedule_path_map.blank?
+
+    schedulable_libraries = library_schedule_path_map.keys
+
+    if schedulable_libraries.one?
+      return link_to_schedule_once_visit(
+        library: schedulable_libraries.first,
+        text: "🗓 Schedule access to #{library_name(schedulable_libraries.first)}",
+        css_class: 'btn btn-primary'
+      )
+    end
+
+    render 'schedules/schedule_library_visit_dropdown', schedulable_libraries: schedulable_libraries
+  end
+
   private
 
   def library_schedule_path_map
-    {
-      'EAST-ASIA' => schedule_eal_path,
-      'GREEN' => schedule_green_path
-    }
+    map = {}
+    map['GREEN'] = schedule_green_path if patron_or_group.can_schedule_green_access?
+    map['EAST-ASIA'] = schedule_eal_path if patron_or_group.can_schedule_eal_access?
+
+    map
   end
 end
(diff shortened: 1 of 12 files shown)


app/models/patron.rb
```diff
@@ -229,6 +229,17 @@ def can_schedule_green_access?
       (profile_key == 'CNAC' && affiliations.include?('affiliate:fellow'))
   end
 
+  def can_schedule_eal_access?
+    return unless Settings.schedule_once.eal_visits.enabled
+
+    faculty = %w[CNF MXF]
```
jkeck, line 235: This _could_ effectively be replaced with `can_schedule_green_access?` since the logic is the same (CodeClimate even complains about this). I wasn't sure if we'll end up getting shifting requirements for access, but I would be happy to refactor this a bit.  My intention is also to not have a dependency on Green being enabled in order for EAL to be enabled which would be the case if we simply relied on the `can_schedule_green_access?` accessor directly.

app/helpers/summaries_helper.rb
```diff
@@ -7,4 +7,40 @@ def scheduler_data_param(user_attributes)
...
+
+  def schedule_once_link_or_dropdown
+    return 'Not eligible during current phase of Research Restart Plan' if library_schedule_path_map.blank?
+
+    schedulable_libraries = library_schedule_path_map.keys
+
+    if schedulable_libraries.one?
+      return link_to_schedule_once_visit(
+        library: schedulable_libraries.first,
+        text: "🗓 Schedule access to #{library_name(schedulable_libraries.first)}",
```
jkeck, line 29: @jvine the 🗓 wasn't in the mockup on #557, but I added it here.  Let me know if I should remove it.

app/views/schedules/_schedule_library_visit_dropdown.html.erb
```diff
@@ -0,0 +1,10 @@
+<div class="schedule-once-dropdown dropdown">
+  <button class="btn btn-primary dropdown-toggle" type="button" id="scheduleOnceDropdown" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+    🗓 Schedule access to ...
```
jkeck, line 3: @jvine the 🗓 wasn't in the mockup on #557, but I added it here. Let me know if I should remove it.

anarchivist: Incidentally, that was probably an artifact from my functional mock (using the old ScheduleOnce embed). @jvine's mocks had another icon.

jvine: Oops yes, it's appropriate to add whatever calendar icon we are using to that button. Missed it in my mock, sorry.

app/models/patron.rb
cbeer: I'm good keeping it separate until we know more/better.

PR #562, opened 2020-06-17
Add academic staff to the list of profiles that can schedule access …

### LINKED ISSUE #561: Add academic staff to eligible group for SPEC and GREEN visits
Academic staff need to be enabled for Special Collections and Green visits.

affiliation data => staff:academic
User Profile => CNAC

jkeck:
…to green, spec, and EAL.

Closes #561

### DIFF
--- a/app/models/patron.rb
+++ b/app/models/patron.rb
@@ -225,8 +225,7 @@ class Patron
     grad_students_and_postdocs = %w[MXD RED REG REG-SUM]
     visiting_scholars = %w[MXAS]
 
-    [*faculty, *grad_students_and_postdocs, *visiting_scholars].include?(profile_key) ||
-      (profile_key == 'CNAC' && affiliations.include?('affiliate:fellow'))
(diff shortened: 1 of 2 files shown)


app/models/patron.rb
```diff
@@ -261,6 +259,12 @@ def can_schedule_special_collections_visit?
 
   private
 
+  def academic_staff_or_fellow?
+    return unless profile_key == 'CNAC'
+
+    affiliations.include?('staff:academic') || affiliations.include?('affiliate:fellow')
```
jkeck, line 265: @saseestone just to validate we're are only taking users w/ profile CNAC and affiliation is `staff:academic` (in addition to `affiliate:fellow`) so we exclude users w/ affiliations like `staff:otherteaching` and `staff:emeritus` (as well as a few other cases)?

saseestone: Please add staff:otherteaching.  Thanks!

PR #568, opened 2020-07-09
Add Pickup Scheduling Support for GSB (via LibCal embed)

### LINKED ISSUE #567: Add pickup scheduling capability for BUSINESS
BUSINESS manages their appointments through LibCal.  We can look into doing a similar iframe embed that we do for ScheduleOnce (but will need to build the hooks in to have different scheduling embeds).

If that doesn't work, we can link to their libcal instance instead.

@jkeck / @anarchivist have the details about their libcal instance and how to embed it (they have a script that injects an iframe and we may want to do the same thing we did w/ ScheduleOnce and just modal the iframe that would have been generated).

### LINKED ISSUE #569: Button wording changes
- [ ] Change "Schedule access to" to "Schedule visit to"
- [ ] Change "Schedule Special Collections visit" to "Schedule visit to Special Collections" (in both its locations)

...for consistency and clearer contrast with "Schedule pickup at"

Screen Shot 2020-07-09 at 3 37 42 PM

jkeck:
Closes #567
Closes #569 

Screen Shot 2020-07-09 at 3 11 53 PM

### DIFF
--- a/app/assets/stylesheets/modules/scheduling.scss
+++ b/app/assets/stylesheets/modules/scheduling.scss
@@ -1,4 +1,4 @@
-.schedule-once-iframe {
+.schedule-iframe {
   box-sizing: border-box;
   display: block;
   margin: 0 auto;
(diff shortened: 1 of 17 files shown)


jkeck:
Button wording changes for #569 

Screen Shot 2020-07-10 at 9 32 35 AM

PR #572, opened 2020-07-18
Update pickup library configs to match the service as it is in …

jkeck:
…requests currently.


config/environments/test.rb
```diff
@@ -45,4 +45,56 @@
 
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
+
```
jkeck, line 48: I prob. should have done this in requests as well (instead of the `if` statements in `application.rb`).

PR #591, opened 2020-09-09
Enable staff to schedule visits

### LINKED ISSUE #590: Add regular non-academic staff, temp/casual staff, and emeritus staff to eligible group for SPEC visits and GREEN and EAL entry
Non-academic (regular) staff and temp-casual staff should now be able to schedule visits to Special Collections, and entry appointments for Green and EAL. 

Regular staff: 
* Authentication privgroup: `stanford:staff` or `stanford:staff:otherteaching`
* SYM user profile: `CNS`, `MXS`, `CNAC`

Temporary staff:
* Authentication privgroup: `stanford:staff:temporary`
* SYM user profile: `MXS`

Casual staff:
* Authentication privgroup: `stanford:staff:casual`
* SYM user profile: `MXS`

jkeck:
Shared configs PRs coming soon

Closes #590

### DIFF
--- a/app/models/patron.rb
+++ b/app/models/patron.rb
@@ -218,30 +218,17 @@ class Patron
     true
   end
 
-  def can_schedule_green_access?
+  def can_schedule_access?(library)
     return if expired?
-    return unless Settings.schedule_once.green_visits.enabled
+    return unless Settings.schedule_access[library]
 
     faculty = %w[CNF MXF]
     grad_students_and_postdocs = %w[MXD RED REG REG-SUM]
     undergrads = %w[REU REU-SUM]
     visiting_scholars = %w[MXAS]
+    staff = %w[CNAC CNS MXAC MXS]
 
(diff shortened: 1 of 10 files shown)


app/models/patron.rb
```diff
@@ -259,9 +246,7 @@ def can_schedule_pickup?(library)
   end
 
   def can_schedule_special_collections_visit?
```
jkeck, line 248: I kept this as-is for the time being (trying to keep the changes to a minimum-ish), but imagine it could get rolled up into one of the methods above.

app/models/patron.rb
```diff
@@ -218,30 +218,17 @@ def can_pay_fines?
     true
   end
 
-  def can_schedule_green_access?
+  def can_schedule_access?(library)
```
jkeck, line 221: `can_schedule_access?` and `can_schedule_access?` now have the same user groups.  One of them could theoretically call the other, or a shared method.  Happy to add that refactor, but stopped just short of doing so.

PR #593, opened 2020-09-15
Make the scheduling modals 900px tall.

jkeck:
per @jvine (based on user feedback)


jkeck:
This is currently deployed to UAT

PR #616, opened 2020-11-09
Use GitHub action instead of Travis for CI.

jkeck:
Note that #610 exists (which I hadn't seen when I started), but this uses the same version of ruby that's on the server and updates the badge.


jgreben:
Thanks @jkeck

PR #628, opened 2021-01-04
Fixes Advisory: CVE-2020-26247

jgreben:


Gemfile
```diff
@@ -44,7 +44,8 @@ gem 'bootstrap', '~> 4.3'
 
 gem 'warden'
 
-gem 'nokogiri'
+# Fixes Advisory: CVE-2020-26247
```
mejackreed, line 47: Can we get away with just doing a `bundle update` with commenting about the CVE or specifying a version? I mention this just as it might become troublesome to try and keep track of that.

mejackreed (changes requested):
Overall looks good to me, just one minor comment.

Gemfile
jgreben: Ok, I updated the Gemfile.lock and removed the version pinning and changed the Gemfile comment.

mejackreed: Great! Looks like there is a Rubocop issue still. I would be fine if you wanted to remove the comment entirely.

mejackreed (approved):
👏

PR #658, opened 2021-05-18
Remove renewal eligibility for items with Item Cat5 of "NORENEW" 

### LINKED ISSUE #649: Remove renewal eligibility for items with Item Cat5 of "NORENEW"
We need to remove the hold override that we're currently passing when a user requests to renew an item that has a title-level hold. We also need to display to the user that they cannot renew the item.  If there is a hold, regardless of the hold level, we should display "No. Another user is waiting for this item."

Background:
The hold override is currently (i.e. in COVID times, when we aren't recalling materials from users) causing a bigger issue than it is meant to solve. It's allowing users to renew materials, even when there is a hold on the item for another user.  In non-COVID times, these title-level holds would convert to Copy level holds as soon as Symphony determined that the checked out copy was the only available copy. That determination happened daily via Symphony report.  In non-COVID times it made sense to error on the side of allowing the renewal for any hold that remained title-level.

But in COVID times, these holds are remaining as Title-level holds for months because Stanford Libraries has chosen not to recall items from users. Even when we go back to recalling materials, there is a concern that we're allowing renewals to address a super edge-case that would occur in a very narrow time-frame.

cbeer:
Fixes #649.

This PR adds a new condition to stop patrons from renewing items that are marked as `NORENEW`. It also adds additional authorization for this condition within the application because Symphony won't enforce it for us.

### DIFF
--- a/app/controllers/renewals_controller.rb
+++ b/app/controllers/renewals_controller.rb
@@ -59,10 +59,15 @@ class RenewalsController < ApplicationController
     params.require(%I[resource item_key])
   end
 
+  # Make sure the checkout belongs to the user trying to do the renewal, and
+  # also make sure the item is not renewable for reasons that symphony doesn't
+  # know about (e.g. the itemCat5 hack we're doing during COVID-19 times; other
+  # conditions are handled by business logic in Symphony so we don't need to
+  # worry about being exhaustive here.)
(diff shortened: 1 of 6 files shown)


camillevilla (approved):
🚢

PR #665, opened 2021-06-18
Rails61 redo

jgreben:
There were conflicts in the Gemfile.lock and deploy problems with the previous `rails61` PR. I checked out a new branch from the updated `main` branch and checked out/added the changed filed from the original `rails61` PR.


jgreben:
@cbeer, your original PR looked fine, but I had some trouble resolving conflicts as noted. I'd like to punt back to you to do the merge if that's OK.

jgreben:
@cbeer This is ready for review. If you want to check out a new branch and re-submit it for my review, that's OK with me. I had to request that node be upgraded on -dev in order for to install all the node_modules, so before deploying this tp uat or prod, we will have to check that off as well.

PR #680, opened 2021-08-24
Send USER_PRIVILEGE_OVRCD when changing a patron's PIN

### LINKED ISSUE #678: changeMyPin action needs USER_PRIVILEGE_OVRCD for staff level client ID
Current web services requests use a client ID that is mapped to the staff level Symphony interface. Web services no longer allows the changeMyPin action to be completed with this staff level interface without an override code sent. Adding `SD-Prompt-Return:USER_PRIVILEGE_OVRCD/XXXX` to the header should make these requests work again.

### LINKED ISSUE #687: do not display appointment when hold is available at pickup library = LAW
Law notified me today that they are re-opening the library to Stanford community users, and they would like to hide the appointment feature that is currently enabled when the user has an available hold at Pickup library = LAW.  We'll need to remove it from view on both the Summary and Requests tabs.

This means we'll only have 1 possible appointment type: for Special Collections visits. The design seems heavy for just one type, with the "Visit Reading Room" button looking disabled for most users most of the time.  But several libraries (including Law) have mentioned they'd want to re-instate pickup appointments IF they have to make drastic changes to library entrance policies. I think this means we live with the current page setup for now.

cbeer:
Fixes #678

### DIFF
--- a/app/services/symphony_client.rb
+++ b/app/services/symphony_client.rb
@@ -121,10 +121,12 @@ class SymphonyClient
   end
 
   def change_pin(token, pin)
-    request('/user/patron/changeMyPin', method: :post, json: {
-      resetPinToken: token,
-      newPin: pin
-    })
+    request(
+      '/user/patron/changeMyPin',
+      method: :post,
+      json: { resetPinToken: token, newPin: pin },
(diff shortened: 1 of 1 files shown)


thatbudakguy:
closing in favor of https://github.com/sul-dlss/mylibrary/pull/702

Issue #571, opened 2020-07-15
Update layout of Feedback form


ggeisler:
As part of the effort to standardize our feedback forms, below are recommendations for updating the My Library Account feedback form. Note that the basic layout updates have been made to Earthworks (sul-dlss/earthworks#691). Also, the update to the placement of the "Check system status" link, which is not included in the Earthworks form, should follow the same approach as recommended in the [SW feedback form update](https://github.com/sul-dlss/SearchWorks/issues/2497).

Here's the current feedback form:

<img width="1112" alt="Screen Shot 2020-07-14 at 4 02 09 PM" src="https://user-images.githubusercontent.com/101482/87599071-42752400-c6a8-11ea-9cbc-587bc2c10b12.png">

---

- [x] A: Add a Bootstrap alert-primary component to the top of the form and move the "Reporting from: current page URL" string to that component.

- [x] B: Move the "Check system status" link into the alert-primary component. It should be right-aligned, or given ~40px left margin from the "Reporting from:" string. The check status link should have font-weight: 600; to draw more attention to it than the neighboring "Reporting from:" link.

I believe MyLibrary is on the same version of Bootstrap as Earthworks, so we should be able to the use the column strategy used there for positioning and sizing the form fields and related elements.

The updated form should look something like:

<img width="991" alt="Screen Shot 2020-07-14 at 4 22 53 PM" src="https://user-images.githubusercontent.com/101482/87600398-c4fde380-c6a8-11ea-86ec-f6cdff589c4f.png">

---

with the version for an authenticated user the same except the Captcha is omitted.

PR #704, opened 2022-01-07
Update layout of feedback form (#571)

thatbudakguy:
Borrows layout from SearchWorks, see:
https://github.com/sul-dlss/SearchWorks/pull/2660

### DIFF
--- a/app/views/shared/feedback_forms/_form.html.erb
+++ b/app/views/shared/feedback_forms/_form.html.erb
@@ -1,10 +1,8 @@
 <div class="container">
   <%= form_tag feedback_form_path, method: :post, class:"feedback-form", role:"form" do %>
-    <div class="row">
-      <div class="col-sm-8 offset-sm-1">
-        <div class="offset-sm-x">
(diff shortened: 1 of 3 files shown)


thatbudakguy:
I've deployed this branch to UAT (https://mylibrary-uat.stanford.edu) so that you can see what the form looks like. I tried to match the image in https://github.com/sul-dlss/mylibrary/issues/571, but SearchWorks seems a bit different from that, so advice is appreciated.

ggeisler (changes requested):
This looks great to me @thatbudakguy.

I haven only the most minor comment, which is that the "Check system status" text is in bold here, while we don't use bold for it in our other applications, I don't think. I don't really mind the bold, but then again, this link isn't the main point of the form and it wouldn't hurt to be consistent across apps, so maybe remove the bold styling?

thatbudakguy:
ok, unbolded the text and redeployed to UAT. not sure if we need another review since it's (theoretically) just frontend code changing, but i'll leave it up for awhile in case anyone else wants to review

camillevilla (approved):

Issue #722, opened 2022-04-19
DRAFT: For 5/2/22: Replace appointment links with short term user message; then remove entirely in early June 2022.


saseestone:
Part of work to open SUL to visitors and Library ID card holders on 5/2/22 - including the removal of appointments for Special Collections and Law Library pickup.

### For 5/2/22, on both Summary tab and in Requests tab

- [ ] Remove links to scheduling software for both:
- Available hold with Pickup library = LAW, resulting in  “Pickup requests at Law Library (Crown)” button. (Link opens LibCal)
- Available hold with Pickup library = SPEC-DESK, resulting in “Visit Reading Room” button. (Link opens ScheduleOnce)
- [ ] Replace schedule links with a short term statement that appointments are no longer required. 
- Text should read “As of May 2, 2022 appointments are no longer needed to enter the Special Collections Reading Room, or pickup materials at the Law Library. Simply arrive after you receive email confirmation that materials are ready for use.”

### For 6/6/22
- [ ] Remove the entire “Schedule appointments” section on the Summary tab
- [ ] Remove statement on Requests tab

saseestone:
@cbeer I'll leave you to place the "design-needed" label if you want Gary to design the short term statement.

saseestone:
@corylown I was testing something else and noted that we have a different statement appearing in the mylibrary Summary tab when a user is ineligible for appointments. 

You can see it by using the fixture user "Fee1" (Library ID = Fee1. PIN = Fee1)

Here's a screengrab of their Summary tab:
![image](https://user-images.githubusercontent.com/14897546/165633276-d78ee2dc-0b04-4de4-b0a1-66c4c4b0799b.png)

For these "ineligible users", we can follow the same plan as for eligible users outlined above (i.e. see the short term statement until 6/6/22. After 6/6/22, the "Schedule appointments" feature can go away.)

corylown:
@saseestone this behavior is controlled by a configuration that had not been yet been applied to the dev server. I applied the configuration change and now the expected temporary message is showing on dev with the Fee1 pin user:

https://mylibrary-dev.stanford.edu/
<img width="891" alt="Screen Shot 2022-04-28 at 8 49 55 AM" src="https://user-images.githubusercontent.com/458247/165755823-ae1a3927-c90f-4c9e-b7fe-31915a673d49.png">

saseestone:
Thanks @corylown it looks beeeeautiful on dev.

PR #726, opened 2022-06-01
Remove scheduling functionality; part of #722

cbeer:


thatbudakguy (approved):
Screen Shot 2022-06-03 at 10 22 53

nice!!

PR #729, opened 2022-07-08
Update to Rails 7

cbeer:


.github/workflows/ruby.yml
```diff
@@ -9,12 +9,15 @@ on:
 jobs:
   tests:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        ruby: ['2.7.1', '3.0', '3.1']
```
thatbudakguy, line 14: is there anything holding this stack back from moving to 3.x only? just wondering

thatbudakguy (approved):

.github/workflows/ruby.yml
cbeer: Just merging this branch and doing the work to get off ruby 2.x

PR #741, opened 2022-08-30
FOLIO updates

thatbudakguy:
- Create FolioClient for FOLIO API requests
- Rework/stub Patron model to use FolioClient
- Add debug FOLIO output to summary page


app/services/cql_query.rb
```diff
@@ -0,0 +1,21 @@
...
+    @args = args
+    @sortby = sortby
+  end
+
+  def to_query
+    [
+      args.map { |k, v| "#{k}==\"#{escape(v)}\"" }.join(' and '),
+      ("sortby #{Array(sortby).join(' ')}" if sortby)
+    ].compact.join(' ').strip
+  end
```
thatbudakguy, line 14: could this just be `to_s`? or could we alias `to_s` to this? i wonder if that will make it not necessary to call explicitly in some places, like the call sites in `FolioClient` currently

marlo-longley:
## 9/23/2022: Status of this branch

- `services/folio-graphql-client.rb` is the datasource. The long query string here is generated from the Apollo interface and copied into the file. It's certain that this query will need to modified in the future to grab all the data we need. If there is a field missing or error in this code, the first place to check is the query string -- make sure you are asking for that field in the query.
- Changing `BibRecord` to `FolioRecord` and using it across all tabs. `BibRecord` seemed to be a file for storing coming accessors across all the tabs. The Requests tab wasn't using it previously; since the data returned from FOLIO for Requests (or "Holds" in FOLIO) now conforms to the other types of data, it made sense to reuse these methods for every tab. I renamed it to `FolioRecord` because `BibRecord` didn't seem accurate.
- `fine.rb`, `checkout.rb`, `request.rb` and `patron.rb`, and the shared `folio_record.rb` models are where most of the alignment needs to be done to match the FOLIO fields. Some comments have been left, and the wiring to FOLIO does work for basic fields like title, author, itemId, etc. But to move forward we will need to be more confident in the mappings from Symphony to FOLIO so that we select the right fields. Anything related to locations, renewals, CDL, circulation rules/notes, patron groups, and other areas are totally left untouched for now, and just set to nil so that the views will load.

cbeer:
rebased into #812

PR #761, opened 2022-12-01
Implement BorrowDirect ReShare requests list for patron.

### LINKED ISSUE #757: For 12/13/22: Replace current BorrowDirect API with new ReShare API for request status.
Copying @saseestone 's note from the sul-requests issue that's related to this:

> IPLC, the governing body for BorrowDirect, is moving away from Relais to ReShare. We will need to replace our current API with a new ReShare API when routing RecallHold requests to BorrowDirect.

> There is API documentation available in basecamp: https://3.basecamp.com/5319802/buckets/26606684/documents/5132521312
Since it requires a basecamp account, I've copied it here: https://docs.google.com/document/d/1u8NouCFH1DVBV74vNtxv30zp9D0HmQI6yGrCMV9W7Wo/edit#
Hilary Thorsen arranged for my basecamp account, and we can request others through her as needed.

> IPLC plans to cut over from Relais to ReShare on ~~12/1/22~~ OR on ~~12/6/22~~ maybe for real on 12/13/22.

mod-rs API will let us lookup request status. We'll need a **username**, **password**, and **tenant** **(do we need to request these from someone?)**.

corylown:
BorrowDirect is changing back-end providers. This adds a client class for making requests to BorrowDirect reshare's new folio based system, a new request model to make the request for a patron's BorrowDirect requests, and wraps the response JSON in a class to mimic the Symphony request object behavior.

Closes #757 

Looks like (no design change from current implementation):

Screen Shot 2022-12-01 at 10 53 58 AM

### DIFF
--- a/app/models/borrow_direct_reshare_requests.rb
+++ b/app/models/borrow_direct_reshare_requests.rb
@@ -1,0 +1,120 @@
+# frozen_string_literal: true
+
+class BorrowDirectReshareRequests
+  attr_reader :patron
+
+  def initialize(patron)
+    @patron = patron
+  end
+
+  def requests
+    # Ensure ReShare requests are active (not yet in Symphony)
+    # and that each ReShare request object's patron_id matches
+    # the logged in patron's university_id.
+    reshare_requests.select(&:active?)
+                    .select { |request| request.patron_id == patron.university_id }
+  end
+
+  private
+
+  def reshare_requests
+    return [] unless patron.university_id
+
+    request_client.requests(patron.university_id).map do |request|
+      ReshareRequest.new(request)
+    end
+  end
+
+  def request_client
+    @request_client ||= BorrowDirectReshareClient.new
+  end
+
+  ##
+  # Wrap the borrow direct reshare request item JSON in a class
+  # so we can give it a similar interface to Symphony Requests
+  class ReshareRequest
+    def initialize(request_json = [])
+      @request_json = request_json
+    end
+
+    # Request becomes REQ_CHECKED_IN once we receive it (and should show as a request ready for pickup/checkout)
+    # Request becomes REQ_REQUEST_COMPLETE once the uesr returns it
+    ACTIVE_REQUEST_STATUSES = %w[
+      REQ_IDLE
+      REQ_VALIDATED
+      REQ_SOURCING_ITEM
+      REQ_SUPPLIER_IDENTIFIED
+      REQ_REQUEST_SENT_TO_SUPPLIER
+      REQ_BORROWING_LIBRARY_RECEIVED
+      REQ_SHIPPED
+      REQ_ERROR
+      REQ_END_OF_ROTA
+      REQ_INVALID_PATRON
+      REQ_EXPECTS_TO_SUPPLY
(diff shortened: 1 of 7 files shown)


app/services/borrow_direct_reshare_client.rb
```diff
@@ -0,0 +1,91 @@
...
+                 password: Settings.borrow_direct_reshare.password,
+                 tenant: Settings.borrow_direct_reshare.tenant)
+    uri = URI.parse(url)
+
+    @base_url = url
+    @username = username
+    @password = password
+
+    if uri.user
+      @username ||= uri.user
```
jcoyne, line 24: Is there a reason we need to handle both the case where the user/pass are encoded in the URI and one where they are passed in?

jcoyne (changes requested):

app/models/borrow_direct_reshare_requests.rb
```diff
@@ -0,0 +1,118 @@
...
+    # and that each ReShare request object's patron_id matches
+    # the logged in patron's university_id.
+    reshare_requests.select(&:active?)
+                    .select { |request| request.patron_id == patron.university_id }
+  end
+
+  private
+
+  def reshare_requests
+    (request_client.requests(patron.university_id) || []).map do |request|
```
jcoyne, line 21: How about having `requests()` return an array (rather than nil)?

app/services/borrow_direct_reshare_client.rb
```diff
@@ -0,0 +1,91 @@
...
+
+  def get(path, **kwargs)
+    authenticated_request(path, method: :get, **kwargs)
+  end
+
+  def get_json(path, **kwargs)
+    parse(get(path, **kwargs))
+  end
+
+  def requests(university_id)
```
jcoyne, line 43: Why do we need to handle the case where there is no university_id?  Should this be an ArgumentError?

app/models/borrow_direct_reshare_requests.rb
corylown: I like it.

app/services/borrow_direct_reshare_client.rb
corylown: I think you're right that this shouldn't be handled here. I'll add a guard clause on BorrowDirectReshareRequests#reshare_requests to handle the case where a patron doesn't have a university_id.

app/services/borrow_direct_reshare_client.rb
```diff
@@ -0,0 +1,91 @@
...
+
+  def authenticated_request(path, headers: {}, **other)
+    request(path, headers: headers.merge('x-okapi-token': session_token), **other)
+  end
+
+  def request(path, headers: {}, method: :get, **other)
+    HTTP.headers(default_headers.merge(headers))
+        .request(method, base_url + path, **other)
+  rescue HTTP::Error => e
+    Honeybadger.notify(e)
```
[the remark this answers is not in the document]
corylown: I was hoping to prevent the user from seeing an error if something went wrong with the reshare request, but I'm open to removing this rescue.

PR #767, opened 2023-01-04
updates dependencies

shelleydoljack:


jgreben (changes requested):
Maybe disable `RSpec/PendingWithoutReason` in rubocop.yml (and whatever else rubocop might be complaining about) and commit anything after running `rubocop -A`.

PR #790, opened 2023-04-04
Migrate analytics to GA4

marlo-longley:
Part of #759

### DIFF
--- a/app/assets/javascripts/analytics.js
+++ b/app/assets/javascripts/analytics.js
@@ -1,190 +1,164 @@
-GoogleAnalytics = (function() {
-  function GoogleAnalytics() {}
+// gtag intial setup
+window.dataLayer = window.dataLayer || [];
+function gtag(){dataLayer.push(arguments);}
+gtag('js', new Date());
 
-  GoogleAnalytics.load = function() {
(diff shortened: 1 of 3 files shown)


marlo-longley:
I guess this fails codeclimate. I can extract some functions to reduce repetition, but not sure there is a ton of benefit in this case. Up to reviewers.

app/assets/javascripts/analytics.js
```diff
@@ -1,190 +1,164 @@
...
+// Send event in required GA4 format to Google
+function sendAnalyticsEvent({ action, category, label, value }) {
+  window.gtag && window.gtag('event', action, {
+    event_category: category,
+    event_label: label,
+    event_value: value
+  });  
+}
+
+document.addEventListener('turbolinks:load', function() {
```
thatbudakguy, line 15: does this fire on the very first page load? like, if I visit the site and do nothing else, will it still run and enable debug mode?

thatbudakguy (approved):

app/assets/javascripts/analytics.js
marlo-longley: @thatbudakguy and I confirmed that the events are sending on the first page load.

Issue #736, opened 2022-08-22
Use the mod-patron (edge-patron?)API to cancel a hold


cbeer:

thatbudakguy:
this ticket is mostly just to confirm the API functions how it says it does and isn't broken.

jcoyne:
It seems to be broken (or the documentation is incorrect)

thatbudakguy:
current status:
- we probably don't want to use edge-patron because it's not installed and that has implications for all patron-related routes
- it's possible to use mod-circulation to do this, but more investigation is needed to see what behavior the ILS actually exhibits with respect to the request and if it's different from using mod-patron to cancel
- we don't want to outright destroy requests since staff sometimes examine previous canceled requests for a patron
- we want to make sure that any workflows tied to cancellations still happen

jcoyne:
It looks like the FOLIO UI just updates the request for a cancel:
```
XHR PUT
https://okapi-test.stanford.edu/circulation/requests/39a4ebd1-0137-45ff-a6a2-c256de260840
[HTTP/1.1 204 No Content 5255ms]

	
cancellationAdditionalInformation	"Seems good to cancel."
cancellationReasonId	"b35c8771-04c6-4eb2-a481-b9480563e84c"
cancelledByUserId	"ad914b1b-06fe-5aef-8446-d60b57395f2f"
cancelledDate	"2023-05-15T19:09:00.216Z"
fulfilmentPreference	"Hold Shelf"
id	"39a4ebd1-0137-45ff-a6a2-c256de260840"
instance	{…}
contributorNames	[…]
0	{…}
name	"Graham, Harry C. (Harry Crusen), 1874-1922"
1	{…}
name	"United States. Department of Agriculture. Bureau of Statistics"
identifiers	[…]
0	
identifierTypeId	"c858e4f2-2b6b-4385-842b-60732ee14abb"
value	"agr13000006"
1	
identifierTypeId	"7e591197-f335-4afb-bc6d-a6d76ca3bace"
value	"(OCoLC-M)4965192"
2	
identifierTypeId	"439bfbae-75bc-4f74-9fc7-b2a2d47ce3ef"
value	"ocm04965192"
publication	[…]
0	{…}
dateOfPublication	"1912"
place	"Washington, D.C"
publisher	"U.S. Dept. of Agriculture, Bureau of Statistics"
title	"Coffee : production, trade, and consumption by countries / by Harry C. Graham."
[...]

Issue #735, opened 2022-08-22
Support renewing material in mylibrary


cbeer:

thatbudakguy:
this ticket is mostly just to confirm the API functions how it says it does and isn't broken.

jcoyne:
I attemped to use the API to renew a record and I get a timeout:
```
irb(main):001:0> client = FolioClient.new
=>
#<FolioClient:0x000000010ba94988
...
irb(main):002:0> user_id = client.lookup_user_id('jcoyne85')
=> "562a5cb0-e998-4ea2-80aa-34ac2b536238"
irb(main):003:0> client.renew_item(user_id, '1a4280f5-a917-589d-8d1c-b6f915667f0e')
/Users/jcoyne85/workspace/sul-dlss/mylibrary/app/services/folio_client.rb:59:in `check_response': Renew request for user_id: 562a5cb0-e998-4ea2-80aa-34ac2b536238, item_id: 1a4280f5-a917-589d-8d1c-b6f915667f0e was not successful. status: 500, io.vertx.core.http.impl.NoStackTraceTimeoutException: The timeout period of 5000ms has been exceeded while executing POST /circulation/renew-by-id for server okapi:9130 (RuntimeError)
```

thatbudakguy:
work merged in https://github.com/sul-dlss/mylibrary/pull/795 but unclear if it's working

corylown:
I think this is working. I was able to renew a couple of items using `client.renew_item` and then got sensible error messages in cases when the item could not be renewed:

**Successful item renewal:**

```
irb(main):006:0> client.renew_item('e8cf1192-7de7-4bb0-880e-3c985f718e96','abe9a6df-b3cf-50c9-a64b-4cdb8b3228c9')
=> nil
```


![Image](https://github.com/sul-dlss/mylibrary/assets/458247/91945ee2-86d9-4cf6-a612-e53f0811a2f8)


```
irb(main):007:0> client.renew_item('e8cf1192-7de7-4bb0-880e-3c985f718e96','abe9a6df-b3cf-50c9-a64b-4cdb8b3228c9')
/Users/corylown/workspace/mylibrary/app/services/folio_client.rb:175:in `check_response': Renew request for user_id: e8cf1192-7de7-4bb0-880e-3c985f718e96, item_id: abe9a6df-b3cf-50c9-a64b-4cdb8b3228c9 was not successful. status: 422, { (RuntimeError)
  "errors" : [ {
[...]

PR #795, opened 2023-05-05
Add FolioClient for renewing items and canceling holds

jcoyne:
Ref #735 #736

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -41,7 +41,9 @@ gem 'honeybadger'
 
 gem 'config'
 
-gem 'http'
+gem 'faraday' # Used by FolioClient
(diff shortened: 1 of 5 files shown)


thatbudakguy:
could we cherry-pick b6a6eb6 onto this as well? It looks like we were worried about escaping things in these CQL queries, and I think it makes sense to bring in `CqlQuery` with the client. I don't think we need anything else from #741

jcoyne:
@thatbudakguy I cherry-picked the Cql class into this pr and moved to use faraday-retry

thatbudakguy (approved):
looks great, thanks!!

Issue #825, opened 2023-05-31
Hook up checkouts display + functionality


cbeer:
This ticket builds on the initial FOLIO mappings + https://github.com/sul-dlss/mylibrary/pull/824. These are the known TODOs:

- [x] `renewable_at` / `renew_from_period` (related to https://github.com/sul-dlss/folio-graphql/issues/34?)
  > folio loan policies have a string field called `renewalsPolicy.renewFromId` described as the "Renew from date". we currently only have one loan policy that has this field defined, and it's set to the string `SYSTEM_DATE`. i don't really know what this means...maybe it's a reference to another field on the loan object itself?
  > for items where renewals have the same length as regular checkouts, this information is an object stored at `loansPolicy.period` on the loan policy:
   ```js
        "loansPolicy": {
         "period": { "duration": 28, "intervalId": "Days" },
        }
     ```
  > 
  > some loan policies can also specify an alternate time length for renewals, which supersedes the length of a regular checkout. if present, that information is in `renewalsPolicy`:
  > 
  ```js
    "renewalsPolicy": {
     "differentPeriod": true,
     "period": { "duration": 5, "intervalId": "Days" },
    }
  ```
  > we don't currently have any policies with a different length for renewals.
  
- [x] `reserve_item?`
  > there's currently no information in a loan policy that identifies it as being specifically for a reserve item (although we do usually put something in the description field about it).
- [x] seen + unseen renewals remaining (FOLIO doesn't make this distinction, so presumably we'll set seen renewals to 0 and use the policies count for unseen renewals)

   > the total number of allowed renewals is renewalsPolicy.numberAllowed on the loan policy.     the number of renewals the patron has used up for that particular item is renewalCount on the checkout object.

- [x] `library`
- [ ] `from_ill?` / `from_ilb?` / `from_borrow_direct?` (https://github.com/sul-dlss/mylibrary/issues/833)
- [x] `renewal_date` / `recalled_date` (https://github.com/sul-dlss/folio-graphql/issues/35)
   > recalls and renewals can use the [/loan-storage/loan-history](https://s3.amazonaws.com/foliodocs/api/mod-circulation-storage/r/loan-storage.html#loan_storage_loan_history_get) API with a CQL query filtering on both the loan ID and the action type.

Actions:
- [x]  confirm we can renew eligible material.

cbeer:
[...]

PR #842, opened 2023-06-08
Set renewals counts for FOLIO

corylown:
Part of #825

### DIFF
--- a/app/models/folio/checkout.rb
+++ b/app/models/folio/checkout.rb
@@ -217,16 +217,16 @@ module Folio
       record.dig('details', 'renewalCount') || 0
     end
 
-    # nil means "unlimited" for unseenRenewalsRemaining
     def unseen_renewals_remaining
-      Float::INFINITY
-      # fields['unseenRenewalsRemaining'] || Float::INFINITY
+      unseen_renewals_allowed - renewal_count
     end
 
-    # nil means "unlimited" for seenRenewalsRemaining
+    def unseen_renewals_allowed
+      record.dig('details', 'loanPolicy', 'renewalsPolicy', 'numberAllowed') || Float::INFINITY
+    end
+
     def seen_renewals_remaining
       Float::INFINITY
-      # fields['seenRenewalsRemaining'] || Float::INFINITY
     end
 
     def library_key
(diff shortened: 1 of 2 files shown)


app/models/folio/checkout.rb
```diff
@@ -217,16 +217,16 @@ def renewal_count
...
+
+    def unseen_renewals_allowed
+      record.dig('details', 'loanPolicy', 'renewalsPolicy', 'numberAllowed') || Float::INFINITY
     end
 
-    # nil means "unlimited" for seenRenewalsRemaining
     def seen_renewals_remaining
-      Float::INFINITY
-      # fields['seenRenewalsRemaining'] || Float::INFINITY
+      0
```
corylown, line 229: The ticket suggests that `seen_renewals_remaining` should be set to 0, but I don't know what seen vs. unseen means so I don't know how to assess whether this is correct.

corylown: We think Infinity is the right default, because: https://github.com/sul-dlss/mylibrary/blob/main/app/models/folio/checkout.rb#L81

app/models/folio/checkout.rb
```diff
@@ -217,16 +217,16 @@ def renewal_count
...
 
-    # nil means "unlimited" for unseenRenewalsRemaining
     def unseen_renewals_remaining
-      Float::INFINITY
-      # fields['unseenRenewalsRemaining'] || Float::INFINITY
+      unseen_renewals_allowed - renewal_count
+    end
+
+    def unseen_renewals_allowed
+      record.dig('details', 'loanPolicy', 'renewalsPolicy', 'numberAllowed') || Float::INFINITY
```
cbeer, line 225: If there isn't a renewals policy, should we default to 0?

```suggestion
      record.dig('details', 'loanPolicy', 'renewalsPolicy', 'numberAllowed') || 0
```

corylown: I guess? I wasn't sure if the nil = unlimited still applied in Folio land: https://github.com/sul-dlss/mylibrary/blob/main/app/models/symphony/checkout.rb#L210

cbeer: 🤔

Issue #827, opened 2023-05-31
Hook up holds display + functionality in mylibrary


cbeer:
TODO:
- [x] `patron_key` (does FOLIO even have a notion of holds for a proxy?)
- [x] `queue_length` 
- [x]  borrowing limit (https://github.com/sul-dlss/folio-graphql/issues/33)
- [x] `from_borrow_direct?`/  `from_ill` (#833)
- [x] stub out CDL stuff so that nothing errors; deferring CDL until after launch

Verify we can:

- [x] cancel a hold (#829)
- [ ] change the pickup library (https://github.com/sul-dlss/mylibrary/pull/840)

marlo-longley:
There is a draft PR for "change the pickup library" at #840 
Waiting for further work on this until we know more about whether the data from the service-points.json (API response) has data we can use, specifically around staff-only pickup locations. But the verification that we can use FOLIO to change a pickup location is complete -- this works in the UI.

marlo-longley:
I just spent a while looking into the `queue_length` property. I couldn't find anything in the FOLIO docs for getting the total queue length. Is it possible we could ship with just the waitlist position? If I am #8 on the waitlist, do I care if I am #8 out of 20? We have the ability to get that the queuePosition is 8.

cbeer:
My memory is... patrons want to be aware that there's other requests after them (because they'll get a reduced checkout period).

marlo-longley:
Gotcha. 

It's possible we could get all the open request for the requested item and add them up, something along the lines of: `circulation/requests?item=#{item_id}&status=open` -- but there isn't a direct endpoint related to the request, from what I can tell

marlo-longley:
I found some quite old documentation that made reference to `circulation/requests/#{request_id}/positions`, where then you could add up all the positions, but that endpoint doesn't work for me on our FOLIO / doesn't exist anymore?

marlo-longley:
Re proxy users placing holds. From the FOLIO docs on adding a Proxy:
https://docs.folio.org/docs/users/#add-a-proxy
> Optional: Select whether the Proxy can request for the sponsor. If you select Yes, then the proxy can place requests for materials on behalf of the sponsor.

It seems like yes, proxies can place requests.

saseestone:
> It's possible we could get all the open requests for the requested item and add them up, something along the lines of: `circulation/requests?item=#{item_id}` and filter for open status -- but there isn't a direct endpoint related to the request, from what I can tell

[...]

PR #856, opened 2023-06-19
Populate patron borrowing limit

marlo-longley:
Part of https://github.com/sul-dlss/mylibrary/issues/827

Screenshot 2023-06-19 at 16 31 02


marlo-longley:
This doesn't address the per-item limits discussed in https://github.com/sul-dlss/folio-graphql/issues/33 -- not sure if that needs to be included here.

Issue #830, opened 2023-06-02
Hook up fees and fines display


cbeer:
Building on https://github.com/sul-dlss/mylibrary/pull/828:

- [x] Translations of FOLIO feeFineTypes into patron-visible labels
~~`sequence` (used to... figure out if the patron recently paid?)~~ broken out into https://github.com/sul-dlss/mylibrary/issues/832
- [x] is `owed` the right field? fixture fines didn't seem to have it
- [x] is `dateCreated` the right field to replace `billDate`? fixture fines didn't seem to have that populated
- [x] should there be an `itemId` attached to fees + fines? the API docs say it's required, but the fixtures didn't seem to have it

thatbudakguy:
there may be fee/fine types in FOLIO that are (always) missing data that we would otherwise expect to have; might be UX implications for handling this

marlo-longley:
re the presence of `itemId` being required. There are currently 3 users with fees:
- undergrad1 and blocked1 have fees with associated barcodes. `itemId` is returned from the API. 
- faculty1 has a fee without a barcode. `itemId` is null. Seemingly this didn't cause problems when creating the record in the UI, but the API throws errors:


![Image](https://github.com/sul-dlss/mylibrary/assets/1328900/841bb877-51f6-41c7-8815-45a0553ed773)

 

![Image](https://github.com/sul-dlss/mylibrary/assets/1328900/059771a9-a7fc-41a0-b11e-0e2e5ea1b055)

marlo-longley:
Re: "translation of FeeFineTypes to patron visible labels" -- These are the current labels we get. They seem OK. @saseestone , do these need editing or changing in any way?

`{
  "data": {
    "feeFineTypes": [
      {
        "id": "9523cb96-e752-40c2-89da-60f3961a488d",
        "feeFineType": "Overdue fine",
      },
      {
        "id": "d20df2fb-45fd-4184-b238-0d25747ffdd9",
        "feeFineType": "Replacement processing fee"
      },
      {
        "id": "cf238f9f-7018-47b7-b815-bb2db798e19f",
        "feeFineType": "Lost item fee",
      },
      {
        "id": "c7dede15-aa48-45ed-860b-f996540180e0",
        "feeFineType": "Lost item processing fee",
      },
      {
        "id": "73785370-d3bd-4d92-942d-ae2268e02ded",
        "feeFineType": "Lost item fee (actual cost)",
      },
      {
[...]

PR #868, opened 2023-06-30
Display proxy activity for checkouts and requests

marlo-longley:
Part of #827

### DIFF
--- a/app/models/folio/checkout.rb
+++ b/app/models/folio/checkout.rb
@@ -106,6 +106,11 @@ module Folio
       record.dig('details', 'proxyUserId') || record.dig('details', 'userId')
     end
 
+    # @return [Boolean] Returns true if the proxyUserId exists
+    def proxy_checkout?
+      record.dig('details', 'proxyUserId').present?
+    end
+
     def overdue?
       record['overdue']
     end
--- a/app/models/folio/group.rb
+++ b/app/models/folio/group.rb
@@ -18,15 +18,9 @@ module Folio
                    end
     end
 
-    def barred?
-      # TODO: decide how this should work in FOLIO
-      false
-    end
+    delegate :barred?, to: :sponsor
 
-    def blocked?
-      # TODO: decide how this should work in FOLIO
-      false
-    end
+    delegate :blocked?, to: :sponsor
 
     def status
       standing
--- a/app/models/folio/group.rb
+++ b/app/models/folio/group.rb
@@ -8,15 +8,24 @@ module Folio
     end
 
     def sponsor
-      nil
+      @sponsor ||= if sponsor?
+                     self
+                   else
+                     # if you are a proxy, make a second call to get the sponsor
+                     # for now, we only support one sponsor per proxy
+                     sponsor_id = user_info.dig('proxiesFor', 0, 'userId')
+                     Folio::Patron.find(sponsor_id)
+                   end
     end
 
     def barred?
-      members.any?(&:barred?)
+      # TODO: decide how this should work in FOLIO
+      false
     end
 
     def blocked?
-      members.any?(&:blocked?)
+      # TODO: decide how this should work in FOLIO
+      false
     end
 
     def status
@@ -34,24 +43,29 @@ module Folio
     end
 
     def checkouts
-      @checkouts ||= member_list.flat_map(&:group_checkouts)
+      sponsor.group_checkouts || []
     end
 
     def fines
-      @fines ||= member_list.flat_map(&:group_fines)
+      sponsor.group_fines
     end
 
     def requests
-      @requests ||= member_list.flat_map(&:group_requests)
+      sponsor.group_requests
     end
 
     def member_list
-      @member_list ||= members.reject { |patron| patron.key == key }
+      @member_list ||= if sponsor?
+                         user_info['proxiesOf']
+                       elsif proxy_borrower?
+                         sponsor.group.member_list
+                       end
     end
 
     def member_list_names
       @member_list_names ||= member_list.each_with_object({}) do |member, hash|
-        hash[member.key] = member.display_name
+        hash[member['proxyUserId']] =
+          "#{member.dig('proxyUser', 'personal', 'firstName')} #{member.dig('proxyUser', 'personal', 'lastName')}"
       end
     end
 
@@ -61,12 +75,6 @@ module Folio
 
     def to_partial_path
       'group/group'
-    end
-
-    private
-
-    def members
-      @members ||= []
     end
   end
 end
--- a/app/models/folio/patron.rb
+++ b/app/models/folio/patron.rb
@@ -12,6 +12,12 @@ module Folio
       @payment_in_process = payment_in_process
     end
 
+    # pass a FOLIO user uuid and get back a Patron object
+    def self.find(key)
+      patron_info = FolioClient.new.patron_info(key)
+      new(patron_info)
+    end
+
     def user_info
       patron_info['user'] || {}
     end
@@ -86,9 +92,11 @@ module Folio
     end
 
     def patron_type
-      patron_group = user_info.dig('patronGroup', 'desc')
+      # FOLIO's patronGroup refers to the patron type, e.g. Undergraduate, Graduate, Faculty, etc.
+      # this type of group is unrelated to our proxy/sponsor "research groups" in the model Folio::Group
+      patron_type = user_info.dig('patronGroup', 'desc')
 
-      return 'Fee borrower' if patron_group.match?(/Fee borrower/i)
+      return 'Fee borrower' if patron_type.match?(/Fee borrower/i)
 
       # suppress the display of any other patron groups
       nil
@@ -107,8 +115,6 @@ module Folio
     end
 
     def display_name
-      return proxy_borrower_name if proxy_borrower_name.present?
-
       "#{first_name} #{last_name}"
     end
 
@@ -125,16 +131,9 @@ module Folio
       borrow_limit - checkouts.length
     end
 
-    def proxy_borrower?
-      false
-    end
-
-    def sponsor?
-      false
-    end
-
-    def checkouts
-      @checkouts ||= patron_info['loans'].map { |checkout| Checkout.new(checkout) }
+    def all_fines
+      # @all_fines ||= []
+      @all_fines ||= patron_info['accounts'].map { |fine| Fine.new(fine) }
     end
 
     def fines
@@ -142,12 +141,13 @@ module Folio
       # all_fines.reject { |fine| payment_sequence.include?(fine.sequence) }
     end
 
-    def all_fines
-      # @all_fines ||= []
-      @all_fines ||= patron_info['accounts'].map { |fine| Fine.new(fine) }
+    # TODO: figure this out for FOLIO
+    def group_fines
+      # return fines.select { |fine| group_billuser_info_keys.include?(fine.key) } if sponsor?
+      # fines
+      []
     end
 
-    ##
     # Creates a range of integers based of a payment sequence string
     def payment_sequence
       return Range.new(0, 0) unless payment_in_process[:billseq] && payment_in_process[:pending]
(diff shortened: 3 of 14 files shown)


app/views/requests/_request.html.erb
```diff
@@ -72,6 +73,10 @@
         <dt class="col-3 offset-1 col-md-2 offset-md-2">Requested:</dt>
         <dd class="col-8"><%= l(request.placed_date, format: :short) %></dd>
       <% end %>
+      <% if request.proxy_requester_name %>
```
cbeer, line 76: Isn't this the same information that we're showing in the "Borrower" `dt` above for Symphony?

marlo-longley: That <dt> is nested under group, which I thought we were getting rid of before today's conversation. I will work on this now that I understand groups better.

marlo-longley:
Thanks for the review @cbeer, the syntax is now changed.

app/models/folio/patron.rb
```diff
@@ -125,76 +132,68 @@ def remaining_checkouts
...
+    def checkouts
+      all_checkouts.reject(&:proxy_checkout?)
     end
 
-    def group_holduser_info_keys
-      @group_holduser_info_keys ||= SymphonyDbClient.new.group_holduser_info_keys(key)
+    # Checkouts from the proxy group
+    def group_checkouts
+      return all_checkouts.select(&:proxy_checkout?) if sponsor?
     end
```
cbeer, line 187:
```suggestion
    def group_checkouts
      all_checkouts.select(&:proxy_checkout?)
    end
```

Does the return value for non-sponsors matter? Or, if it does, could we return an empty array to simplify any downstream consumers?

app/models/folio/request.rb
```diff
@@ -26,7 +27,12 @@ def key
     end
 
     def patron_key
-      record.dig('details', 'userId')
+      record.dig('details', 'proxyUserId') || @patron_id
```
cbeer, line 30:
```suggestion
      record.dig('details', 'proxyUserId') || record.dig('details', 'userId')
```

Do we need to pass through the original patron? I thought the FOLIo request would have that information somewhere already.

marlo-longley: I kept getting null for non proxies. I was confused about this. Doing this made it work but perhaps too heavy a solution / not the right fix.

marlo-longley: At the very least, there is no such thing as `details, userId` in a request
Screenshot 2023-07-12 at 16 40 35

app/models/folio/patron.rb
marlo-longley: group_checkouts is only ever called in the `Group` class itself by `Group#checkouts`

```
   def checkouts
      sponsor.group_checkouts
    end
```

So I think the `if sponsor?` return is good. I will add that back.

app/models/folio/request.rb
cbeer: `requesterId`, maybe?

app/models/folio/patron.rb
cbeer: I can't help but foresee a situation where a proxy user is also a sponsor and everything breaks. If it's fine to just return `all_checkouts.select(&:proxy_checkout?)` that seems best, but if there's some reason that'd be bad, at least returning `[]` would keep things from throwing exceptions

app/models/folio/request.rb
marlo-longley: Oof.

app/models/folio/patron.rb
marlo-longley: I think it would be fine. This is consumed in the `checkouts_controller`:
```
    @checkouts = patron_or_group.checkouts.sort_by { |x| x.sort_key(:due_date) }
```

I could add the safe operator here and I think we'd be all good?
```
    @checkouts = patron_or_group.checkouts&.sort_by { |x| x.sort_key(:due_date) }
```

saseestone: Thankfully only Faculty members are eligible - by longstanding policy - to have proxies.  I'm not saying it's never going to happen that a proxy is also a sponsor...but it shouldn't happen.

app/controllers/checkouts_controller.rb
```diff
@@ -9,7 +9,7 @@ class CheckoutsController < ApplicationController
   # GET /checkouts
   # GET /checkouts.json
   def index
-    @checkouts = patron_or_group.checkouts.sort_by { |x| x.sort_key(:due_date) }
+    @checkouts = patron_or_group.checkouts&.sort_by { |x| x.sort_key(:due_date) }
```
jcoyne, line 12: Is there a reason we have the nil check here?  Doesn't `#checkouts` always return an array? If not, can you make it always return an array?

app/models/folio/checkout.rb
```diff
@@ -106,6 +106,11 @@ def patron_key
       record.dig('details', 'proxyUserId') || record.dig('details', 'userId')
     end
 
+    def proxy_checkout?
+      # return true if proxyUserId exists
```
jcoyne, line 110: Would you mind moving this comment before the method and perhaps use YARD doc (e.g. `# @returns true if proxyUserId exists?`)

app/models/folio/group.rb
```diff
@@ -8,16 +8,17 @@ def patron_type
     end
 
     def sponsor
-      nil
-    end
+      return self if sponsor?
```
jcoyne, line 11: Shouldn't this be cached in @sponsor too? Then you don't have to repeat this check many times?

app/models/folio/group.rb
```diff
@@ -34,24 +35,29 @@ def standing
...
-      @requests ||= member_list.flat_map(&:group_requests)
+      sponsor.group_requests
     end
 
     def member_list
-      @member_list ||= members.reject { |patron| patron.key == key }
+      if sponsor?
+        @member_list ||= user_info['proxiesOf']
+      elsif proxy_borrower?
+        sponsor.group.member_list
```
jcoyne, line 53: Is there a reason this doesn't also get cached in @member_list?

app/models/folio/patron.rb
```diff
@@ -30,6 +36,7 @@ def university_id
 
     def status
       if proxy_borrower?
+        # TODO: I'm not sure about group statuses anymore in FOLIO
```
jcoyne, line 39: I don't think this is a helpful comment.  What are you not sure about?

app/models/folio/patron.rb
```diff
@@ -125,76 +132,68 @@ def remaining_checkouts
...
-    end
-
-    def sponsor?
-      false
-    end
-
-    def checkouts
-      @checkouts ||= patron_info['loans'].map { |checkout| Checkout.new(checkout) }
+    def all_fines
+      # @all_fines ||= []
```
jcoyne, line 136: Cruft?

app/models/folio/patron.rb
```diff
@@ -125,76 +132,68 @@ def remaining_checkouts
...
 
     def fines
       all_fines
       # all_fines.reject { |fine| payment_sequence.include?(fine.sequence) }
     end
 
-    def all_fines
-      # @all_fines ||= []
-      @all_fines ||= patron_info['accounts'].map { |fine| Fine.new(fine) }
+    # TODO: figure this out for FOLIO
```
jcoyne, line 145: Is there a ticket for this yet?

app/models/folio/request.rb
```diff
@@ -26,7 +26,12 @@ def key
     end
 
     def patron_key
-      record.dig('details', 'userId')
+      record.dig('details', 'proxyUserId') || record.dig('details', 'requesterId')
+    end
+
+    def proxy_request?
+      # return true if proxyUserId exists
```
jcoyne, line 33: Can you move this before the method and use YARD doc?

spec/models/folio/request_spec.rb
```diff
@@ -0,0 +1,81 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Folio::Request do
+  subject(:request) do
+    described_class.new(record.with_indifferent_access)
+  end
+
+  # rubocop:disable Layout/LineLength
```
jcoyne, line 10: Can you add returns on line 18, 23-24, 30 instead? Long lines are hard to read.

spec/models/folio/request_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  describe '#key' do
+    it 'has a request key' do
+      expect(request.key).to eq '572919e2-0817-49df-87bc-04c9775ae48d'
+    end
+  end
+
+  describe '#patron_key' do
+    context 'when the request is not a proxy request' do
```
jcoyne, line 51: Can you set `subject(:patron_key) { request.patron }` key here.

spec/models/folio/request_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+      'canceledByUserId' => nil,
+      'cancellationAdditionalInformation' => nil,
+      'canceledDate' => nil,
+      'patronComments' => nil
+    }
+    # rubocop:enable Layout/LineLength
+  end
+
+  describe '#key' do
+    it 'has a request key' do
```
jcoyne, line 45: Can you set `subject(:key) { request.key }` key here.

spec/models/folio/request_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+            'proxy' => { 'firstName' => 'Piper', 'lastName' => 'Proxy', 'barcode' => 'Proxy1' } } }
+      end
+
+      it { expect(request.patron_key).to eq 'bdfa62a1-758c-4389-ae81-8ddb37860f9b' }
+    end
+  end
+
+  describe '#proxy_request?' do
+    context 'when the request is not a proxy request' do
+      it { expect(request.proxy_request?).to be false }
```
jcoyne, line 68: Use `expect(request).not_to be_proxy_request`

spec/models/folio/request_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+
+    context 'when the request is a proxy request' do
+      let(:record) do
+        { 'details' =>
+          { 'proxyUserId' => 'bdfa62a1-758c-4389-ae81-8ddb37860f9b',
+            'proxy' => { 'firstName' => 'Piper', 'lastName' => 'Proxy', 'barcode' => 'Proxy1' } } }
+      end
+
+      it { expect(request.proxy_request?).to be true }
```
jcoyne, line 78: Use `expect(request).to be_proxy_request`

jcoyne (changes requested):

app/controllers/checkouts_controller.rb
marlo-longley: I put this here to address Chris's comment above

jcoyne: I don't see the comment you are referring to.

app/models/folio/patron.rb
marlo-longley: All these comments among fines were existing before this PR. They were commented out to make the model work in a first pass. The only reason this is coming up as a change is because I moved it within the file. They will be fixed in https://github.com/sul-dlss/mylibrary/issues/830

app/models/folio/patron.rb
marlo-longley: https://github.com/sul-dlss/mylibrary/issues/830 - yes

app/controllers/checkouts_controller.rb
marlo-longley: Chris asked:
> Does the return value for non-sponsors matter? Or, if it does, could we return an empty array to simplify any downstream consumers?

And his code suggestion didn't include the returned empty array. So I tried to follow that. I can add the array back but this seemed to go on the track of the return value not mattering for non-sponsors.

marlo-longley:
I addressed the review comments and added a lot more testing.

jcoyne (approved):

Issue #833, opened 2023-06-02
How do we know if an item is from ILL/ILB/BorrowDirect/etc?


cbeer:

saseestone:
Asked Hilary in Resource Sharing about this. She wasn't sure, but promised to provide us some exemplar records soon (i.e. next week)

thatbudakguy:
we know we have some of these (lots, actually) in folio-test, but they're suppressed from discovery and not loaned out currently. @saseestone to work on the data some more

saseestone:
So the "lots of these in folio-test" are migrated items that Darsi brought over from Symphony into folio-test. Luckily, she didn't migrate the loans or holds on these records, so in theory, I can commandeer some for our testing.

When I tried to create checkouts and requests, I'm getting a "no loan policy" message so we've asked the circ folks to create the correct circrules for these items.

marlo-longley:
Thanks for looking into it @saseestone !

saseestone:
Ok checkouts and pages created on migrated borrowdirect inventory items. Users are undergrad1 (page that is ready for pickup) and faculty1 (one checkout, one page ready for pickup).

marlo-longley:
We can use location codes within `effectiveLocation`. Here is an example of data returned in folio-graphql for a hold placed on a BorrowDirect item:

``` json    
     {
        "details": {
          "itemId": "2a47e79d-12e7-5f62-ab96-c3e85a864546",
          "instanceId": "edcdba3e-0d4f-5b3e-9e0e-37f5c2ecbad8",
          "tags": null
        },
        "item": {
          "item": {
            "effectiveLocation": {
              "code": "SUL-BORROW-DIRECT",
              "description": null,
              "name": "SUL Borrow Direct"
            }
          },
          "instance": {
            "id": "edcdba3e-0d4f-5b3e-9e0e-37f5c2ecbad8"
          }
        }
      }
```
So in this case, we have `SUL-BORROW-DIRECT` as the effectiveLocation code.
Currently there are no codes for ILL or ILB. But we can use this same approach when those codes are created.

With [this PR](https://github.com/sul-dlss/mylibrary/pull/863), we will have that info available for both holds and loans (was already retrieved as part of the loan info).

PR #869, opened 2023-07-01
Use location code to determine if a request is from BorrowDirect

marlo-longley:
Based on #833

Wanted to get this up while I was on the topic. However, I'm not sure of any current requests with this code to test with in the UI and will keep this as a draft until I test that.

### DIFF
--- a/app/models/folio/request.rb
+++ b/app/models/folio/request.rb
@@ -207,9 +207,12 @@ module Folio
       nil
     end
 
+    def location_code
+      record.dig('item', 'item', 'effectiveLocation', 'code')
+    end
+
     def from_borrow_direct?
-      # TODO
-      nil
+      location_code == 'SUL-BORROW-DIRECT'
     end
 
     def item_type
--- a/config/application.rb
+++ b/config/application.rb
@@ -62,7 +62,6 @@ module Mylibrary
       'ARS' => 'Archive of Recorded Sound',
       'ART' => 'Art & Architecture Library (Bowes)',
       'BIOLOGY' => 'Biology Library (Falconer)',
-      'BORROW_DIRECT' => 'BorrowDirect',
       'BUSINESS' => 'Business Library',
       'CDL' => 'Digital lending',
       'CHEMCHMENG' => 'Chemistry & ChemEng Library (Swain)',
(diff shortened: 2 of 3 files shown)


app/models/folio/request.rb
```diff
@@ -208,8 +204,7 @@ def from_ilb?
     end
 
     def from_borrow_direct?
-      # TODO
-      nil
+      library_key == 'SUL-BORROW-DIRECT'
```
cbeer, line 207: I thought this would probably be a location code, not a whole library?

marlo-longley:
@cbeer this should make more sense now, thank you.

app/models/folio/request.rb
```diff
@@ -99,7 +99,7 @@ def placed_library
 
     def library
       if from_borrow_direct?
-        Settings.BORROW_DIRECT_CODE
```
cbeer, line 102: Shouldn't we keep `Settings.BORROW_DIRECT_CODE` to make sure we trigger the right behaviors in the view?

marlo-longley: I'm not sure. I thought we wanted to stop using the Settings.

cbeer: We're currently using `BORROW_DIRECT` to get the right name mapping here:

https://github.com/sul-dlss/mylibrary/blob/main/config/application.rb#L65

Are we ready to get those mappings from FOLIO now?

marlo-longley: Ah. Thank you I didn't know this.

marlo-longley: Would it be appropriate to add "SUL-BORROW-DIRECT" as a mapping and make a note to remove the old one after FOLIO?

cbeer: If it were me, I'd keep using the `Settings` value and post-FOLIO we can just use the FOLIO discovery labels.

config/application.rb
```diff
@@ -90,6 +89,9 @@ class Application < Rails::Application
       'SCIENCE' => 'Science Library (Li and Ma)',
       'SPEC-COLL' => 'Special Collections',
       'SPEC-DESK' => 'Special Collections',
+      # TODO: remove this first entry BORROW_DIRECT after FOLIO. SUL-BORROW-DIRECT is the new FOLIO code.
```
jcoyne, line 92:
```suggestion
      # TODO: remove this first entry BORROW_DIRECT after migrating off of Symphony. SUL-BORROW-DIRECT is the new FOLIO code.
```

jcoyne (changes requested):

config/application.rb
marlo-longley: I will add this. The ILL code too will be replaced but can leave off that comment until we are addressing that particular code.

marlo-longley:
@jcoyne I added your suggestion for the comment.

jcoyne:
@marlo-longley I see you changed `settings.yml`, but I asked for the change in `config/application.rb`.

jcoyne (approved):

jcoyne:
@marlo-longley please fix rubocop error and then feel free to merge.

jcoyne (approved):

PR #874, opened 2023-07-14
Stub out CDL logic for FOLIO Request model

marlo-longley:
I commented out the logic so we would have it as reference in the future but I could also just delete it.

Part of #827


jcoyne (changes requested):
I think the code we're not using should be removed. We can always look in git if we need to see what it used to look like.

marlo-longley:
@jcoyne thanks. I removed it.

jcoyne (approved):

PR #878, opened 2023-07-18
Pay fees and fines using CyberSource SecureAcceptance code.

jgreben:
The Payments controller test works, but I am unable to get the other tests to pass, at least locally.


thatbudakguy:
closing in favor of #890; thanks for all the material here that went into that!

PR #882, opened 2023-07-19
Support display of payment history from FOLIO

### LINKED ISSUE #832: Display billing transaction history for a patron
From `Symphony::Payment`:
- [x] `key` (billNumber)
- [x] `bill_description` / `nice_bill_description`
- [x] `bill_amount`
- [x] `bill_date`
- [x] `item_title`
- [x] `item_library`
- [x] `payment_amount`
- [x] `payment_date`
- [x] `resolution`
- [x] `paid_fee?`

thatbudakguy:
Closes #832


app/models/symphony/patron.rb
```diff
@@ -223,6 +223,12 @@ def can_pay_fines?
       true
     end
 
+    def payments
```
marlo-longley, line 226: If we are merging new code related to the Symphony client, should there be a new test?

thatbudakguy: this (and `symphony_legacy_client`) came over from the `SymphonyClient`, so it's more of a move than new code imo. I just needed it to be here instead of on the client, so that `patron.payments` worked for both Symphony and FOLIO

marlo-longley (approved):

PR #887, opened 2023-07-25
Support retaining our auth cookie during payment

thatbudakguy:
- Set the SameSite attribute to None for authentication cookie
- Force SSL when RAILS_ENV=production

This fixes the existing behavior where the user is signed out and redirected to the root after making a payment, because once we go out to Cybersource we lose our authentication cookie.

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -62,4 +62,11 @@ Rails.application.configure do
 
   # Annotate rendered view with file names.
   # config.action_view.annotate_rendered_view_with_filenames = true
+
+  # Chrome (including headless) blocks cookies with SameSite=None (which we set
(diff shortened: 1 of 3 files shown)


config/environments/test.rb
```diff
@@ -62,4 +62,9 @@
 
   # Annotate rendered view with file names.
   # config.action_view.annotate_rendered_view_with_filenames = true
+
+  # Chrome (including headless) blocks cookies with SameSite=None (which we set
```
jcoyne, line 66: Can you add something about Chrome requiring `None ; Secure`, which we are unable to do in test.

thatbudakguy: done

jcoyne (approved):

Issue #889, opened 2023-07-26
Use Cybersource external payment to resolve FOLIO fines


thatbudakguy:
depends on #883. once Cybersource POSTs back to us that the fines have been paid, we need to use FOLIO's bulk pay API to update the `Account`s in the ILS.

- [x] make `FolioClient#pay_fines` mark all of the user's fines as paid
- [x] update tests for `PaymentsController` to test payments for both Symphony and Folio
- [x] make `Folio::Patron#payment_sequence` a no-op and add a TODO about removing it post-launch
- [x] make `Folio::Account#sequence` a no-op and add a TODO about removing it post-launch
- [x] add TODOs about removing payment cookie-related methods post-launch

PR #890, opened 2023-07-27
Handle Cybersource payments without redirecting to Symphony

### LINKED ISSUE #883: Migrate CyberSource payment handling from LibSys scripts
Currently, payment handling is done via some PHP that lives on the Morison VM and communicates with CyberSource, our payment processor, on behalf of MyLibrary. We want to move this implementation into MyLibrary so that when the Symphony VMs go away, payment still works.

thatbudakguy:
This PR supersedes (and was based on) #878. It closes #883.

See the changes to the README for some background on the payment process.

Before this PR, the process of sending information to and from Cybersource via POST was done by some PHP scripts that are located on the Symphony VMs. After the transaction with Cybersource, the PHP scripts in turn called a perl script on the same Symphony VM called `transactionPost.pl` which used some of Symphony's locally-available scripts (not web services or APIs) to actually mark the fine as paid.

After this PR, the code that sends users to and from Cybersource has moved into MyLibrary as ruby (it takes essentially the same approach of rendering an interstitial form). After the transaction is completed, MyLibrary calls the `transactionPost.pl` script to mark the fine as paid, in the same way that it would make other calls to Symphony web services/APIs. This seam allows us to separate the checkout process from the ILS, so that #889 is just a matter of implementing `FolioClient#pay_fines`.

I tested this approach and successfully paid the fines (in Symphony) for user GRAD1, who no longer has any fines to pay. @jgreben confirmed that `transactionPost.pl` was successfully called from MyLibrary. If you want to test the process as well, ask me for the settings that correspond to the "local dev" Cybersource profile, which will redirect you to `localhost` after payment is completed.

config PRs for new Cybersource config:
- https://github.com/sul-dlss/shared_configs/pull/2022
- https://github.com/sul-dlss/shared_configs/pull/2023
- https://github.com/sul-dlss/puppet/pull/9737

not sure whether we need configs for mylibrary-dev (going away) or mylibrary-uat (just use prod config?)

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -80,6 +80,8 @@ en:
     fine_payment:
       accept_html: <span class="font-weight-bold">Success!</span> $%{amount} paid. A receipt has been sent to the email address associated with your account. Payment may take up to 5 minutes to appear in your payment history.
       cancel_html: <span class="font-weight-bold">Payment canceled.</span> No payment was made &mdash; your payable balance remains unchanged.
+      request_failed_html: <span class="font-weight-bold">Sorry!</span> Something went wrong, possibly due to a connection failure. Please <a href="mailto:sul-privileges@stanford.edu">contact us</a> for help resolving your fines.
+      payment_failed_html: <span class="font-weight-bold">Sorry!</span> Payment failed. Please <a href="mailto:sul-privileges@stanford.edu">contact us</a> for help resolving your fines.
     sessions:
       login_by_library_id:
         alert: Unable to authenticate.
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # MyLibrary
+
 [![Build Status](https://github.com/sul-dlss/mylibrary/workflows/CI/badge.svg?branch=main)](https://github.com/sul-dlss/mylibrary/actions?query=workflow%3ACI+branch%3Amain)
 [![Code Climate](https://codeclimate.com/github/sul-dlss/mylibrary/badges/gpa.svg)](https://codeclimate.com/github/sul-dlss/mylibrary)
 [![Test Coverage](https://api.codeclimate.com/v1/badges/a8f1c5dab3a53ffba586/test_coverage)](https://codeclimate.com/github/sul-dlss/mylibrary/test_coverage)
@@ -8,6 +9,7 @@ renews materials, cancels hold requests, makes payments of library fees and fine
 and other features enhancing library user experience.
 
 ## Requirements
+
 1. Ruby (3.1 or greater)
 2. Bundler
 3. Connection to Symphony Web Services (hosted on symws-prod.stanford.edu and symws-dev.stanford.edu)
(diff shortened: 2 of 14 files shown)


.rubocop.yml
```diff
@@ -100,6 +100,10 @@ RSpec/DescribeClass:
 RSpec/VerifiedDoubleReference:
   Enabled: false
 
+RSpec/MultipleMemoizedHelpers:
+  Exclude:
```
jcoyne, line 104: How about just setting a reasonable Max value instead of excluding specific files?

app/controllers/concerns/cybersource.rb
```diff
@@ -0,0 +1,192 @@
...
+    end
+
+    # Comma-separated parameter version of unsigned_fields
+    def unsigned_field_names
+      @unsigned_fields.join(',')
+    end
+
+    # Hash of signed fields and their values
+    def signed_data
+      @signed_fields.index_with { |field| send(field) }
```
jcoyne, line 106: Should this be memoized?  If you call `obj.sign!; obj.to_h` it's called two times. I'd think after signing, we wouldn't want to produce this value again.

app/controllers/concerns/cybersource.rb
```diff
@@ -0,0 +1,192 @@
...
+    def currency
+      'USD'
+    end
+  end
+
+  # Response sent back by Cybersource after the external checkout is complete
+  class PaymentResponse
+    def initialize(fields)
+      @fields = fields
+      @signed_fields = fields['signed_field_names']&.split(',')
```
jcoyne, line 156: Can you explain in which situations `signed_field_names` would not be present or nil?

app/views/payments/cybersource_form.erb
```diff
@@ -0,0 +1,15 @@
+<html charset="utf-8">
```
jcoyne, line 1: Shouldn't this filename end with `.html.erb`

config/locales/en.yml
```diff
@@ -80,6 +80,8 @@ en:
     fine_payment:
       accept_html: <span class="font-weight-bold">Success!</span> $%{amount} paid. A receipt has been sent to the email address associated with your account. Payment may take up to 5 minutes to appear in your payment history.
       cancel_html: <span class="font-weight-bold">Payment canceled.</span> No payment was made &mdash; your payable balance remains unchanged.
+      request_failed_html: <span class="font-weight-bold">Sorry!</span> Something went wrong, possibly due to a connection failure. Please contact us for help resolving your fines.
+      payment_failed_html: <span class="font-weight-bold">Sorry!</span> Payment failed. Please contact us for help resolving your fines.
```
jcoyne, line 84: Should there be directions for contacting us?

app/controllers/concerns/cybersource.rb
```diff
@@ -0,0 +1,192 @@
+# frozen_string_literal: true
+
+# Utilities for generating and validating payment parameters for Cybersource
+module Cybersource
```
jcoyne, line 4: I don't love this as a module. Most of this is business logic, not about a controller. Can we just add the two methods to the Payments controller, and move the other classes to `app/models` or `app/services` as appropriate?

jcoyne (changes requested):

.rubocop.yml
marlo-longley: I set this to 15 in my recent PR

thatbudakguy: cool, I'll rebase and see if that helps

app/controllers/concerns/cybersource.rb
thatbudakguy: Cybersource always signs the response, so I don't think that's possible...something would have to have gone very wrong, and I think it'd manifest in a different way if so.

app/controllers/concerns/cybersource.rb
thatbudakguy: yeah, I was holding off memoizing until I finished testing, but this seems like a good candidate. will do

app/views/payments/cybersource_form.erb
thatbudakguy: oh yeah, good catch!

config/locales/en.yml
thatbudakguy: Yeah, I should make that a link, probably

app/controllers/concerns/cybersource.rb
thatbudakguy: I struggled to find out how to map this to Railsisms like concern, model, etc. and opted to do this just to keep everything in one place. I'm OK moving the two methods to the controller, but what should happen to the rest? Maybe `Security` is a service, and the `PaymentRequest` and `PaymentResponse` are models?

app/controllers/concerns/cybersource.rb
jcoyne: In that case can we drop the safe-navigation operator from this line and the next?

app/controllers/concerns/cybersource.rb
jcoyne: Yes, I like that. You could continue to keep them in the Cybersource namespace too.

app/controllers/concerns/cybersource.rb
thatbudakguy: Sure, that makes sense to me

app/views/payments/cybersource_form.erb
thatbudakguy: fixed

app/controllers/concerns/cybersource.rb
thatbudakguy: done

app/controllers/concerns/cybersource.rb
thatbudakguy: memoized

.rubocop.yml
thatbudakguy: seems like that did it.

app/controllers/concerns/cybersource.rb
thatbudakguy: ok, extracted

jcoyne (approved):

PR #897, opened 2023-07-31
Set SameSite cookie attribute to 'Lax' in development

thatbudakguy:
See #887; this just applies the same fix we currently use in test
to the dev server so that authentication, etc. will work in
local dev.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -43,7 +43,7 @@ Metrics/CyclomaticComplexity:
 Metrics/BlockLength:
   Exclude:
     - 'app/views/**/*.json.jbuilder'
-    - 'config/environments/test.rb'
+    - 'config/environments/*.rb'
     - 'config/routes.rb'
     - 'spec/**/*'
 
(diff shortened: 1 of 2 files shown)


.rubocop.yml
```diff
@@ -43,9 +43,9 @@ Metrics/CyclomaticComplexity:
 Metrics/BlockLength:
   Exclude:
     - 'app/views/**/*.json.jbuilder'
-    - 'config/environments/test.rb'
+    - 'config/environments/*.rb'
     - 'config/routes.rb'
-    - 'spec/**/*'
+    - 'spec/**/*'gs
```
jcoyne, line 48: This is not legal

jcoyne (changes requested):

.rubocop.yml
thatbudakguy: stray keys from my alias; good catch

thatbudakguy: fixed

jcoyne (approved):

PR #901, opened 2023-08-03
Cybersource payments for FOLIO fines

thatbudakguy:
- Add Cybersource dev doc PDF to README and comments
- Disable payment cookie related methods for Folio and add TODOs
- Implement payment of fines for FOLIO
- Handle the case where Cybersource does not send any unsigned fields


app/models/cybersource/payment_response.rb
```diff
@@ -2,13 +2,14 @@
...
   # Response sent back by Cybersource after the external checkout is complete
+  # See the PDF linked in the README for more information on the fields
   class PaymentResponse
     class PaymentFailed < StandardError; end
 
     def initialize(fields)
       @fields = fields
       @signed_fields = fields['signed_field_names'].split(',')
-      @unsigned_fields = fields['unsigned_field_names'].split(',')
+      @unsigned_fields = fields['unsigned_field_names']&.split(',') || []
```
thatbudakguy, line 12: Cybersource always signs _some_ fields, but may elect to sign _all_ of them, in which case `unsigned_field_names` is apparently not sent. TIL

spec/controllers/payments_controller_spec.rb
thatbudakguy, line 1: hiding whitespace helps with this diff

app/models/cybersource/payment_response.rb
jcoyne: That would be a great thing to put in a comment in the code.

thatbudakguy: Added.

jcoyne (approved):

PR #905, opened 2023-08-04
Use service points for FOLIO request options

### LINKED ISSUE #875: Use FOLIO Service Points for request pickup locations
A Draft PR started here https://github.com/sul-dlss/mylibrary/pull/840 but this was before all the Service Points work so it may need to be thrown out. This ticket pulled out from #827.

What needs to change are the options we currently see when trying to modify a request:

Screenshot 2023-07-14 at 13 01 08

marlo-longley:
The service point model code started based on the same model from `sul-requests`

Closes #875

### DIFF
--- a/app/controllers/requests_controller.rb
+++ b/app/controllers/requests_controller.rb
@@ -39,7 +39,7 @@ class RequestsController < ApplicationController
     flash[:success] = []
     flash[:error] = []
 
-    handle_pickup_change_request if params['pickup_library'].present?
+    handle_pickup_change_request if params['service_point'].present?
     handle_not_needed_after_request if params['not_needed_after'].present? &&
                                        params['not_needed_after'] != params['current_fill_by_date']
 
@@ -78,7 +78,9 @@ class RequestsController < ApplicationController
   def handle_pickup_change_request
     change_pickup_response = ils_client.change_pickup_library(*change_pickup_params)
     case change_pickup_response.status
-    when 200
+    # The FOLIO API returns 204
+    # TODO: after FOLIO launch remove the 200 case which was the Symphony response
+    when 200, 204
       flash[:success].push(t('mylibrary.request.update_pickup.success_html', title: params['title']))
     else
       Rails.logger.error(change_pickup_response.body)
@@ -102,7 +104,12 @@ class RequestsController < ApplicationController
   end
 
   def change_pickup_params
-    params.require(%I[resource id pickup_library])
+    # TODO: after Folio launch remove conditional wrapper and delete the else (Symphony) part
+    if ils_client.is_a?(FolioClient)
+      params.require(%I[id service_point])
+    else
+      params.require(%I[resource id service_point])
+    end
   end
 
   def not_needed_after_params
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -65,12 +65,13 @@ module ApplicationHelper
     Mylibrary::Application.config.library_map[code] || code
   end
 
-  # FOLIO distinguishes between service points and libraries (that can share codes)
-  #   so we need a separate method for correct code to name lookup
+  # TODO: After FOLIO launch remove the conditional wrapper and rename to reflect service point terminology
   def pickup_location_name(code)
-    Mylibrary::Application.config.library_map[code] ||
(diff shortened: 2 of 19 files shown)


app/helpers/application_helper.rb
```diff
@@ -65,12 +65,14 @@ def library_name(code)
...
   end
 
-  # FOLIO distinguishes between service points and libraries (that can share codes)
-  #   so we need a separate method for correct code to name lookup
+  # TODO: After FOLIO launch remove the conditional wrapper and rename to reflect service point terminology
   def pickup_location_name(code)
-    Mylibrary::Application.config.library_map[code] ||
-      Folio::ServicePoint.find_by(id: code)&.discoveryDisplayName ||
-      code
+    # debugger
```
jcoyne, line 70: Remove this please

app/controllers/requests_controller.rb
```diff
@@ -78,7 +78,7 @@ def cdl_waitlist_position
   def handle_pickup_change_request
     change_pickup_response = ils_client.change_pickup_library(*change_pickup_params)
     case change_pickup_response.status
-    when 200
+    when 200, 204
```
jcoyne, line 81: Do we need both of these? When does it return one over the other?

marlo-longley: Yes, the folio response is a 204. I can leave a comment about this.

app/helpers/application_helper.rb
marlo-longley: Thx. Done.

app/services/folio_client.rb
```diff
@@ -135,19 +137,21 @@ def cancel_hold_request(user_id, hold_id)
...
   #                                        pickup_location_id: 'bd5fd8d9-72f3-4532-b68c-4db88063d16b')
   # @param [String] hold_id the UUID of the FOLIO hold
-  # @param [String] pickup_location_id the uuid of the new location
-  def change_pickup_location(hold_id:, pickup_location_id:)
-    request_data = get_json("/circulation/requests/#{hold_id}")
-    request_data['pickupServicePointId'] = 'bd5fd8d9-72f3-4532-b68c-4db88063d16b'
-    response = put("/circulation/requests/#{hold_id}", json: request_data)
-
+  # @param [String] service_point_id the uuid of the new location
+  def change_pickup_library(id, service_point)
```
jcoyne, line 147: These parameters no longer match the names of the parameters in the method documentation.

jcoyne (changes requested):
One more small concern and then looks good.

app/helpers/requests_helper.rb
```diff
@@ -5,12 +5,29 @@ module RequestsHelper
...
-          Mylibrary::Application.config.library_map[code]
-        end
-    )
+    # TODO: after FOLIO launch remove this conditional wrapper
+    if request.is_a?(Folio::Request)
+      return options_for_select(
+        service_points(request),
+        # the second param here pre-selects the current service point in the dropdown
+        request.service_point_id
+      )
```
thatbudakguy, line 14: for a function interface like this, I will sometimes mix positional and keyword params. so it would look like:
```rb
def options_for_select(service_points, selected: nil)
```
then when calling, the interface is a little clearer:
```rb
return options_for_select(service_points(request), selected: request.service_point_id)
```
i realize this PR doesn't have anything to do with that function, but it might be worth putting a `TODO: ` to refactor it later.

thatbudakguy (approved):
this looks great; the tests in `requests_helper_spec` were easy to follow and tell the story.

app/helpers/requests_helper.rb
marlo-longley: Thanks @thatbudakguy ! Idk if you knew this was a built-in Rails method -- they define it as you suggest:
https://apidock.com/rails/v5.2.3/ActionView/Helpers/FormOptionsHelper/options_for_select

So I guess here I could add the keyword in to my call.

thatbudakguy: I did not know that! If it's that easy and supports both APIs, then I guess go ahead and do the keyword version. Thanks!

marlo-longley:
Thank you @jcoyne and @thatbudakguy ! Both comments are now addressed.

jcoyne (approved):

PR #934, opened 2023-08-29
Use mod-circulation to cancel requests because mod-patron cannot cancel title level holds

### LINKED ISSUE #932: [mylibrary/prod] FolioClient::IlsError: Cancel request for user_id: 96f1c204-469d-46d5-8cd6-e96ce5d95bcf, hold_id: c0ce3c3a-d7ed-58b1-9338-7ffc1a0a9a68 was not successful. status: 500,
## Backtrace

    line 237 of [PROJECT_ROOT]/app/services/folio_client.rb: check_response
    line 135 of [PROJECT_ROOT]/app/services/folio_client.rb: cancel_hold_request
    line 117 of [PROJECT_ROOT]/app/services/folio_client.rb: cancel_hold

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/62116/faults/99775031)

corylown:
Fixes #932

### DIFF
--- a/app/services/folio_client.rb
+++ b/app/services/folio_client.rb
@@ -118,20 +118,17 @@ class FolioClient
   end
 
   # Cancel a hold request
-  # See https://s3.amazonaws.com/foliodocs/api/mod-patron/p/patron.html#patron_account__id__hold__holdid__cancel_post
   # @param [String] user_id the UUID of the user in FOLIO
   # @param [String] hold_id the UUID of the FOLIO hold
   def cancel_hold_request(user_id, hold_id)
-    # You would think FOLIO could look up this information and merge it itself, but no. You'd
-    # also think you could get the /circulation/requests/{id} endpoint data and use that,
-    # but the API schema is slightly different and OKAPI complains... so we have to look it
-    # up in the patron account data instead.
-    request_data = patron_account(user_id)['holds'].find { |h| h['requestId'] == hold_id }
+    # We formerly used the mod-patron API to cancel hold requests, but it is
+    # unable to cancel title level hold requests.
+    request_data = get_json("/circulation/requests/#{hold_id}")
     request_data.merge!('cancellationAdditionalInformation' => 'Canceled by mylibrary',
-                        'canceledByUserId' => user_id,
-                        'canceledDate' => Time.now.utc.iso8601,
+                        'cancelledByUserId' => user_id,
+                        'cancelledDate' => Time.now.utc.iso8601,
                         'status' => 'Closed - Cancelled')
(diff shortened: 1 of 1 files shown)


app/services/folio_client.rb
```diff
@@ -118,20 +118,17 @@ def cancel_hold(resource, _item_key, patron_key)
...
   # Cancel a hold request
-  # See https://s3.amazonaws.com/foliodocs/api/mod-patron/p/patron.html#patron_account__id__hold__holdid__cancel_post
   # @param [String] user_id the UUID of the user in FOLIO
   # @param [String] hold_id the UUID of the FOLIO hold
   def cancel_hold_request(user_id, hold_id)
-    # You would think FOLIO could look up this information and merge it itself, but no. You'd
-    # also think you could get the /circulation/requests/{id} endpoint data and use that,
-    # but the API schema is slightly different and OKAPI complains... so we have to look it
-    # up in the patron account data instead.
-    request_data = patron_account(user_id)['holds'].find { |h| h['requestId'] == hold_id }
```
cbeer, line 129: After this change, is there anywhere we're making sure the current patron actually has the hold (and didn't just input some other UUIDs into the URL params)?

corylown: Doesn't look like we're validating this anywhere else. Do you think it's okay do something like `return unless request_data['requesterId'] == user_id` within that method? It'll blow up but that seems preferable to cancelling a request for another user, and I think is roughly equivalent to what would have happened with the previous logic. The better thing maybe would be check in the controller and return a 403 if there's a mismatch between the requesterId and the user_id, but requires some an additional api call to get the request and validate it.

corylown: Elected to pass an empty hash in place of the request JSON to the cancel request call to FOLIO if the user id on the request doesn't match the current user.

app/services/folio_client.rb
```diff
@@ -118,20 +118,21 @@ def cancel_hold(resource, _item_key, patron_key)
...
+    # We formerly used the mod-patron API to cancel hold requests, but it is
+    # unable to cancel title level hold requests.
+    request_data = get_json("/circulation/requests/#{hold_id}")
+
+    # Ensure this is the user's request before trying to cancel it
+    request_data = {} unless request_data['requesterId'] == user_id
```
cbeer, line 129: I hate to even ask but... what about proxy requests?

corylown: Looks like the proxy is tracked on the request via `proxyUserId`. I'm curious now to see whether this would have worked for the previous implementation with mod-patron because requests made on a behalf of a sponsor aren't returned on a proxy's account info.

cbeer: Ah, maybe it didn't. Should we file that away for later?

Issue #943, opened 2023-09-01
Alert PIN users that they have to reset their PIN in FOLIO


thatbudakguy:
can we improve on the existing messaging/UI:

![Image](https://github.com/sul-dlss/mylibrary/assets/4924494/c64cb496-7205-4987-a0c3-600f3988e617)

marlo-longley:
Was reading Chris's comment in #946 and thought this could help -- I recall a conversation with @saseestone where we wondered about placing the messaging close to the button impacting these users (the Log in with PIN button) rather than as a global/top level alert. Not sure if that's helpful and there may be obvious UX reasons not to do this!

It would be awesome to have a design that covers both the pages screenshotted below. A process we have used in the past is to post a quick mockup of a potential design here in the ticket, get Sarah's OK and some eyes from the team here in the ticket thread, have any discussion of pros/cons here in the ticket if needed, and then remove the design-needed tag once we're all good. At that point we would consider writing code and/or reassigning the ticket to a developer.

I am totally open to other processes but I thought it could be helpful to explain a general way we've worked in the past (or at least my impression of it). If this is redundant knowledge apologies!

![Screenshot 2023-09-07 at 11 42 18](https://github.com/sul-dlss/mylibrary/assets/1328900/b6fb1abf-048e-4df1-9c08-6772d0d7d69a)
![Screenshot 2023-09-07 at 11 42 28](https://github.com/sul-dlss/mylibrary/assets/1328900/0b1e3133-c0f2-442d-81a5-b38a85e3600f)

alundgard:
Thank you @marlo-longley ! This is very helpful feedback. I really appreciate it. Especially, thanks for explaining the team's common processes and practices. That's great for me to know, and I'm more than happy to contribute in whatever way works best for the team. :-)

And yes, I agree with your and Sarah's design intuition. After getting more context from Chris, I also thought a global alert is maybe not intervening at the right level. Makes sense to put the PIN-specific content beneath the Proxy header. I'll post some mockups today.

alundgard:
Here are some mockups of a PIN message and placement (cc @saseestone @cbeer @marlo-longley ). Feedback welcome!
<img width="1624" alt="Screenshot 2023-09-08 at 12 44 40 PM" src="https://github.com/sul-dlss/mylibrary/assets/2343316/455f1fd8-8ce6-4057-be3b-3e797ff4669e">
[...]

PR #946, opened 2023-09-05
Update global alert style

alundgard:
**Before**: See #943
**After**:
Screenshot of alert banner, now emerald green.

**Changes:** Borrows the emerald green banner color from SUL (to distinguish alerts from the Stanford cardinal).
Screenshot of SUL website, showing emerald green alert banner.

Feedback welcome. Happy to make additional changes (e.g., messaging changes in [global_alerts](https://github.com/sul-dlss/global_alerts)). Wasn't sure how much of a re-design is desired here, but this at least makes the alert a bit more noticeable. Thanks!

### DIFF
--- a/app/assets/stylesheets/modules/global-alert.scss
+++ b/app/assets/stylesheets/modules/global-alert.scss
@@ -1,7 +1,16 @@
- $global-alert-color: rgb(204, 0, 0);
+$global-alert-color: rgb(204, 0, 0);
+$su-bg-digital-green: rgb(0 133 102);
 
 .global-alert {
-  border: 1px solid $global-alert-color;
   margin: 10px auto;
-  padding: 10px;
+  padding: 15px;
+
+  color: white;
+  background: $su-bg-digital-green;
+  border-radius: 8px;
+  font-size: 1.215em;
+
+  p:last-child {
+    margin-bottom: 0;
+  }
 }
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/modules/global-alert.scss
```diff
@@ -1,7 +1,16 @@
- $global-alert-color: rgb(204, 0, 0);
+$global-alert-color: rgb(204, 0, 0);
+$su-bg-digital-green: rgb(0 133 102);
```
jcoyne, line 2: This doesn't seem to match the brand guide: https://identity.stanford.edu/design-elements/color/web/  Digital green should be `rgb(0 155 118)`, right?

jcoyne (changes requested):

app/assets/stylesheets/modules/global-alert.scss
```diff
@@ -1,7 +1,16 @@
- $global-alert-color: rgb(204, 0, 0);
+$global-alert-color: rgb(204, 0, 0);
+$su-bg-digital-green: rgb(0, 155, 118);
```
jcoyne, line 2: It looks like we have all the colors in https://github.com/sul-dlss/mylibrary/blob/main/app/assets/stylesheets/sul-variables.scss  Would it make sense to move these there?

alundgard: Yes, it'd be nice to have all the SUL colors in one place. I noticed those and thought that perhaps they are out-of-date? For example, "mint" (#009b76) appears to correspond to Digital Green (#008566). But, white text on "mint" is not WCAG AA compliant.

cbeer:
> Wasn't sure how much of a re-design is desired here, but this at least makes the alert a bit more noticeable.

My impression of #943 is it wasn't so much about making the alert noticeable (as it only applies to a tiny fraction of users), but rather thinking through how to add the right messaging in the right places for the users it impacts.

PR #962, opened 2023-09-21
Update tests for FOLIO

### LINKED ISSUE #956: Update tests to test against Folio classes instead of Symphony

corylown:
Closes #956 and let's us start removing/refactoring Symphony-isms

### DIFF
--- a/app/services/folio_client.rb
+++ b/app/services/folio_client.rb
@@ -150,12 +150,14 @@ class FolioClient
   # TODO: after FOLIO launch rename this method to reflect service point terminology (maybe change_pickup_service_point)
   #
   # Change hold request service point
-  # @example client.change_pickup_library(hold_id: '4a64eccd-3e44-4bb0-a0f7-9b4c487abf61',
+  # @example client.change_pickup_library(resource: '4a64eccd-3e44-4bb0-a0f7-9b4c487abf61',
+  #                                        _item_key: '4a64eccd-3e44-4bb0-a0f7-9b4c487abf61',
   #                                        pickup_location_id: 'bd5fd8d9-72f3-4532-b68c-4db88063d16b')
-  # @param [String] hold_id the UUID of the FOLIO hold
+  # @param [String] resource the UUID of the FOLIO hold
+  # @param [String] _item_key this was the item key in Symphony
   # @param [String] service_point the UUID of the new service point
-  def change_pickup_library(hold_id, service_point)
-    update_request(hold_id, { 'pickupServicePointId' => service_point })
+  def change_pickup_library(resource, _item_key, service_point)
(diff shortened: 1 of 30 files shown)


app/services/folio_client.rb
```diff
@@ -150,12 +150,14 @@ def cancel_hold_request(user_id, hold_id)
...
   #                                        pickup_location_id: 'bd5fd8d9-72f3-4532-b68c-4db88063d16b')
-  # @param [String] hold_id the UUID of the FOLIO hold
+  # @param [String] resource the UUID of the FOLIO hold
+  # @param [String] _item_key this was the item key in Symphony
   # @param [String] service_point the UUID of the new service point
-  def change_pickup_library(hold_id, service_point)
-    update_request(hold_id, { 'pickupServicePointId' => service_point })
+  def change_pickup_library(resource, _item_key, service_point)
+    update_request(resource, { 'pickupServicePointId' => service_point })
   end
```
thatbudakguy, line 161: would it make sense to change the name of this method in this PR as noted in the comments, and also get rid of the (now totally unused) `item_key`?

spec/controllers/fines_controller_spec.rb
```diff
@@ -3,18 +3,18 @@
 require 'rails_helper'
 
 RSpec.describe FinesController do
-  let(:mock_patron) { instance_double(Symphony::Patron, group?: false, barcode: '1234') }
+  let(:mock_patron) { instance_double(Folio::Patron, group?: false, barcode: '1234') }
 
   let(:fines) do
     [
-      instance_double(Symphony::Fine, key: '1', sequence: '1', owed: '12', status: 'BADCHECK')
+      instance_double(Folio::Account, key: '1', sequence: '1', owed: '12', status: 'BADCHECK')
```
thatbudakguy, line 10: might be another thing that doesn't need to go in this PR, but i believe 'sequence' is a symphony-specific thing because we needed to submit this weird "billing sequence" param to the ILS, which we no longer do.

thatbudakguy (approved):

thatbudakguy:
I see now you noted that this PR is a prelude to actually resolving/changing symphony-isms, so I'm fine to merge it and then afterward address the comments here, fwiw

app/services/folio_client.rb
corylown: I want to handle any refactoring around method names and arguments in future work. This PR is big enough and I think it'll be easier to review the refactoring work if it's approached in small chunks of work.

corylown:
> I see now you noted that this PR is a prelude to actually resolving/changing symphony-isms, so I'm fine to merge it and then afterward address the comments here, fwiw

Yeah -- I think it'll be easier to address any refactoring once this is merged. Then we'll have reasonable tests against FOLIO and can approach the work methodically.

Issue #961, opened 2023-09-21
[mylibrary/prod] FolioClient::IlsError: Renew request for user_id: 2ce1a5eb-8fe4-4487-acd5-5e8ea6901bf6, item_id: 816a1fe2-3661-570d-a1b7-30ff67110a77 was not successful. status: 422, {
  "errors" : [ {
    "message" : "loan at maximum renewal number",
    "parameters" : [ {
      "key" : "loanPolicyName",
      "value" : "1yearfixed-2renew-7daygrace"
    }, {
      "key" : "loanPolicyId",
      "value" : "ad47ac4d-8305-4039-8ac4-93a69a789074"
    } ],
    "overridableBlock" : {
      "name" : "renewalBl...


cbeer:
## Backtrace

    line 256 of [PROJECT_ROOT]/app/services/folio_client.rb: check_response
    line 113 of [PROJECT_ROOT]/app/services/folio_client.rb: renew_item_request
    line 89 of [PROJECT_ROOT]/app/services/folio_client.rb: renew_item

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/62116/faults/100509416)

corylown:
The loan policy on this item changed from `1qtr-3renew-7daygrace` when it was checked out to `1yearfixed-2renew-7daygrace`. This patron already has 2 renewals. I can see how much of a performance hit it is to lookup the latest loan policy for each checked out item.

PR #973, opened 2023-10-06
Implement Folio::Types to keep cached local copies of infrequently changed data from Folio

corylown:
This is some setup to support fixing https://github.com/sul-dlss/mylibrary/issues/961 where we need to find the latest applicable loan policy for a checkout. It implements `Folio::Types` to cache local copies of frequently accessed but relatively stable data from FOLIO and brings mylibrary into more alignment with sul-requests and SearchWorks. I've created a [jenkins pipeline](https://sul-ci-prod.stanford.edu/job/Continuous%20Deployment/job/mylibrary-folio-types-import/) that is currently off that can be enabled once this is merged to keep the cached files up to date.


marlo-longley:
This code looks good to me. It's similar to code in our other apps. Once it is merged, I can also verify the Jenkins is working right. 
I wanted to run through the UI as a final step in this code review but was unable to, since I can't on my current laptop (need to tunnel/ssh and can't without kinit). If anyone else wants to take that on, we could get this merged; otherwise tomorrow Cory and I will try to screenshare and do it together.

PR #989, opened 2023-10-18
Refactor to get library name from FOLIO instead of app config

corylown:


hudajkhan (approved):
Tried it out and appears to work.  The central structure is that library name/location related methods have been moved into record_location, which is included in folio_record (which is also included directly into requests and checkout). Record_location is directly included in account.

Issue #300, opened 2019-07-29
Integration with ILLIAD


cbeer:
Patrons currently use sulils.stanford.edu to view the status of their ILL requests, but Symphony and BorrowDirect requests show up on the Requests tab. This is confusing for our users because we route their hold/recall requests to one of those services automatically.

We want to show both ILL + Scan requests in MyLibrary. 

TODO:
- [ ] Get some fixture data for our test users (or actually ILL some stuff?)
- [ ] Determine how to use ILLIAD `TransactionStatus` field to filter the displayed results (from @saseestone)

Cory did some investigation in #787 and determined:

> It appears that user-initiated requests in Illiad are made via an account that matches that users SUNET ID. We can make a request to the Illiad API for requests from a particular user:
> https://{ILLIAD_API_URL}/ILLiadWebPlatform/Transaction/UserRequests/{SUNET_ID}

> How the request information displayed in MyLibrary could map to the Illiad API response:
> 
> - **Deliver to:** Stanford (hard coded) or `ItemInfo4` (has library code e.g. GREEN)
> - **Not needed after:** in transit (hard coded) (`NotWantedAfter` is also available, but hard-coded is parity)
> - **Title:** `LoanTitle`
> - **Author:** `LoanAuthor`
> - **Requested:** `CreationDate` or `TransactionDate`
> - **Source:** ILL (currently BorrowDirect (hard coded) or `SystemID` (has values like `Reshare:stanford`))
> - **Call Number:** `CallNumber` (mostly this is unpopulated for Borrow Direct requests that I found)

If we can show users all the relevant ILL information, we can also remove (or adjust) the links directing uses to sulils: https://github.com/sul-dlss/mylibrary/blob/main/app/views/requests/index.html.erb#L76-L85

---

Here's what ILLIAD shows for a scan:

![Image](https://user-images.githubusercontent.com/111218/234956385-5eb48eba-9cd3-4dd6-903c-a964785ca9b4.png)
![Image](https://user-images.githubusercontent.com/111218/234956386-8deb69fc-893e-4211-923d-abc8671f919b.png)

saseestone:
@hudajkhan I finally got a chance to write up some documentation for ILLiad & MyAccount, and I hope it proves useful...

https://docs.google.com/document/d/143O12M4HpvJPVbxdKE094KyNNwa3SE16JrBRm1nbRlw/edit.

PR #991, opened 2023-10-18
Integrate Illiad requests into requests page

hudajkhan:
Closes https://github.com/sul-dlss/mylibrary/issues/300 

What this pull request does:

- Request ILLIAD transactions and include them in the list of requests associated with a patron. (These are all self requests and don't fall under any patron group related activities).
- Show a "manage" link for ILLIAD transactions (instead of the change/cancel links) that links to the specific ILLIAD page for that request
- ILLIAD transactions don't show any waitlist information
- Use different fields for a scan ILLIAD transaction vs. a hold/recall ILLIAD transaction
- Have the UI enable sorting in the same way as other regular requests
- Added related tests


app/models/illiad_requests.rb
```diff
@@ -0,0 +1,121 @@
...
+                   [author, title, call_number]
+                 when :call_number
+                   [call_number]
+                 end
+      sort_key.join('---')
+    end
+    # rubocop:enable Metrics/MethodLength
+
+    def date_sort_key
+      (expiration_date || END_OF_DAYS).strftime('%FT%T')
```
corylown, line 67:
```suggestion
      (expiration_date || Folio::Request::END_OF_DAYS).strftime('%FT%T')
```

spec/controllers/requests_controller_spec.rb
```diff
@@ -61,6 +61,21 @@
       end
     end
 
+    describe 'Illiad requests' do
```
corylown, line 64: I don't think you need this test, because you're combining all the request types in the patron model.

spec/models/illiad_requests_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+    before do
+      stub_request(:get, "#{Settings.sul_illiad}ILLiadWebPlatform/Transaction/UserRequests/#{patron_sunet_id}")
+        .to_return(status: 200, body: transaction_results, headers: {})
+    end
+
+    let(:request_results) { ill_request.requests }
+
+    it 'correctly returns request objects' do
+      expect(request_results.length).to be(2)
+    end
```
corylown, line 42:
```suggestion
    it 'correctly returns request objects' do
      expect(ill_request.requests.length).to eq 2
    end
```

spec/models/illiad_requests_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    it 'correctly retrieves title for scan transaction' do
+      expect(ill_scan_request.title).to eq('String Beans')
+    end
+
+    it 'correctly retrieves author for scan transaction' do
+      expect(ill_scan_request.author).to eq('Frederick Wright')
+    end
+  end
```
corylown, line 77:
```suggestion
  describe 'IlliadRequests::Request' do
    context 'when the request is a hold' do
      let(:subject) do
        IlliadRequests::Request.new(JSON.parse(hold_recall_result))
      end

      it 'correctly identifies hold recall result as not being type scan' do
        expect(subject.scan_type?).to be(false)
      end
  
      it 'correctly retrieves title' do
        expect(subject.title).to eq('Disney Motion Pictures')
      end
  
      it 'correctly retrieves author' do
        expect(subject.author).to eq('Gilbert Roy')
      end
  
    end

    context 'when the request is scan' do
      let(:subject) do
        IlliadRequests::Request.new(JSON.parse(scan_result))
      end

      it 'correctly identifies scan result as type scan' do
        expect(subject.scan_type?).to be(true)
      end
  
      it 'correctly retrieves title' do
        expect(subject.title).to eq('String Beans')
      end
  
      it 'correctly retrieves author' do
        expect(subject.author).to eq('Frederick Wright')
      end
    end
```

corylown (changes requested):

spec/controllers/requests_controller_spec.rb
hudajkhan: I added this b/c they had a similar BorrowDirectRequests construct above. I can remove it if not necessary or relevant.

corylown: Yeah, now I see it. Seems like an odd test to me. It's testing whether the thing you assigned in the test is assigned.

hudajkhan: I can remove it

corylown (approved):

PR #1025, opened 2024-01-10
Upgrade to rails 7.1

jcoyne:


config/environments/production.rb
```diff
@@ -33,42 +30,49 @@
...
   # Use a real queuing backend for Active Job (and separate queues per environment).
-  # config.active_job.queue_adapter     = :resque
+  # config.active_job.queue_adapter = :resque
   # config.active_job.queue_name_prefix = "mylibrary_production"
 
   config.action_mailer.perform_caching = false
 
-  # Make sure URL helpers in emails work
-  config.action_mailer.default_url_options = { host: Settings.mailer_host, protocol: 'https' }
-
```
corylown, line 71: I think we want to retain this setting

corylown (changes requested):
Needs rebase

config/environments/production.rb
jcoyne: Got it.

corylown (approved):

PR #1033, opened 2024-01-26
Rename "Library ID" to "University ID"

mjgiarlo:
Connects to sul-dlss/SearchWorks#3773

This commit helps us move the access portfolio beyond using library barcode numbers, leaning instead on the university-assigned and -managed "university ID."

### DIFF
--- a/app/services/folio_client.rb
+++ b/app/services/folio_client.rb
@@ -40,6 +40,12 @@ class FolioClient
   end
 
   def login(university_id, pin)
+    user_response = get_json('/users', params: { query: CqlQuery.new(barcode: university_id).to_query })
+    if (user = user_response.dig('users', 0))
+      return unless validate_patron_pin(user['id'], pin)
+
+      return user
+    end
     user_response = get_json('/users', params: { query: CqlQuery.new(externalSystemId: university_id).to_query })
     user = user_response.dig('users', 0)
 
--- a/app/services/folio_client.rb
+++ b/app/services/folio_client.rb
@@ -39,28 +39,53 @@ class FolioClient
     "#<#{self.class.name}:#{object_id}  @base_url=\"#{base_url}\">"
   end
 
-  def login(university_id, pin)
-    user_response = get_json('/users', params: { query: CqlQuery.new(barcode: university_id).to_query })
-    if (user = user_response.dig('users', 0))
-      return unless validate_patron_pin(user['id'], pin)
-
-      return user
-    end
-    user_response = get_json('/users', params: { query: CqlQuery.new(externalSystemId: university_id).to_query })
-    user = user_response.dig('users', 0)
-
-    return unless user && validate_patron_pin(user['id'], pin)
-
-    user
+  # Login by barcode or university ID, trying barcode first
+  # TODO: remove once we're no longer using barcodes for auth
+  def login_by_barcode_or_university_id(barcode_or_id, pin)
+    login_by_barcode(barcode_or_id, pin) || login_by_university_id(barcode_or_id, pin)
   end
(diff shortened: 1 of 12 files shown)


config/settings.yml
```diff
@@ -47,6 +47,10 @@ analytics_debug: true
 
 mailer_host: 'localhost'
 
+university_id:
+  label: 'University ID'
```
jcoyne, line 51: Can we move these two things to i18n?

jcoyne (changes requested):
Looks pretty good. I think it would be better if we put these labels in i18n though.

config/settings.yml
mjgiarlo: Will do.

mjgiarlo: Done.

jcoyne (approved):

app/services/folio_client.rb
```diff
@@ -176,16 +200,6 @@ def change_pin(token, new_pin)
...
-  # Look up a patron by barcode and return a Patron object
-  # If 'patron_info' is false, don't run the full patron info GraphQL query
-  def find_patron_by_barcode(barcode, patron_info: true)
-    response = get_json('/users', params: { query: CqlQuery.new(barcode:).to_query })
-    user = response.dig('users', 0)
-    raise ActiveRecord::RecordNotFound, "User with barcode #{barcode} not found" unless user
-
-    patron_info ? Folio::Patron.find(user['id']) : Folio::Patron.new({ 'user' => user })
-  end
-
```
thatbudakguy, line 188: Could we keep this, make `find_patron_by_university_id` do what it says, and then have a new method  that tries the other two methods in sequence? That's clearer to me than having `find_patron_by_university_id` the way it is in this PR

thatbudakguy: (I can make this change if others agree)

mjgiarlo: 👍🏻

cbeer: That's fine with me too.. I was trying to leave the code in a "just take these few lines away later" state, but since it sounds like later may be... much later... clarifying it is good too 🤷‍♂️

app/services/folio_client.rb
```diff
@@ -39,22 +39,53 @@ def inspect
...
+  end
+
+  # Find a Folio::Patron by barcode; fetch full patron info if patron_info is true
+  def find_patron_by_barcode(barcode, patron_info: true)
+    user = find_user_by_barcode(barcode)
+    patron_info ? Folio::Patron.find(user['id']) : Folio::Patron.new({ 'user' => user })
   end
 
-  def user_info(user_id)
-    get_json("/users/#{CGI.escape(user_id)}")
```
thatbudakguy, line 57: as far as I could tell, this was unused.

app/services/folio_client.rb
thatbudakguy: Updated to do this.

app/services/folio_client.rb
```diff
@@ -224,6 +245,30 @@ def locations
...
+  end
+
+  # Find a user by sunetid (username in FOLIO); raise an error if not found
+  def find_user_by_sunetid(sunetid)
+    user = get_json('/users', params: { query: CqlQuery.new(username: sunetid).to_query }).dig('users', 0)
+    raise ActiveRecord::RecordNotFound, "User with username '#{sunetid}' not found" unless user
+
+    user
+  end
+
```
thatbudakguy, line 271: these were repeated across the `login_by_*` and `find_patron_by_*` methods, so I extracted them.

PR #1042, opened 2024-02-12
Align settings with sul-requests

marlo-longley:
To reduce the burden of context switching, I am aligning these settings to match what we have in `sul-requests`. I also feel this is just clearer.

See https://github.com/sul-dlss/sul-requests/blob/72d50b864e2c69f941e82b8d29334069c82d176d/config/settings.yml#L65

Since these rely on env vars, I checked `shared_configs`, and we shouldn't need to edit anything there.

If approved I can let everyone know that their local settings files will need to be edited.


config/settings.yml
```diff
@@ -14,10 +14,9 @@ symws:
   login_params: {}
   public_clientID:
 folio:
-  url: <%= ENV['OKAPI_URL'] %>
+  okapi_url: <%= ENV['OKAPI_URL'] %>
   online_service_point_id: 7c5abc9f-f3d7-4856-b8d7-6712462ca007
-folio_graphql:
-  url: <%= ENV['FOLIO_GRAPHQL_URL'] %>
+  graphql_url: <%= ENV['FOLIO_GRAPHQL_URL'] %>
```
jcoyne, line 19: The GraphQL isn't really part of Folio.  I feel like this is maybe a bit misleading and not as clear as what we had before.

dnoneill: If we change it back we should update sul-requests to match mylibrary.

marlo-longley: Thanks for reviewing. I hear your point. I thought it would be pretty nice if our clients and settings looked as similar as possible between the two apps. I would be willing to change sul-requests to go the other way?

PR #1109, opened 2024-06-14
Add stimulus and use it to display a loading message for bulk renewals

corylown:
This also disables the renew all button to prevent duplicate POST requests resulting in error messages from FOLIO.


corylown:
Took a different approach: https://github.com/sul-dlss/mylibrary/pull/1113/files

PR #1117, opened 2024-06-18
Gracefully handle the nil patron_info

### LINKED ISSUE #1115: [mylibrary/prod] NoMethodError: undefined method `[]' for nil:NilClass
## Backtrace

    line 21 of [PROJECT_ROOT]/app/models/folio/patron.rb: user_info
    line 55 of [PROJECT_ROOT]/app/models/folio/patron.rb: barred?
    line 7 of [PROJECT_ROOT]/app/views/shared/_navigation.html.erb: block in _app_views_shared__navigation_html_erb__3823995230013490758_56700

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/62116/faults/101449941)

jcoyne:
Fixes #1115


spec/models/folio/patron_spec.rb
```diff
@@ -16,6 +16,23 @@
...
+      allow(Honeybadger).to receive(:notify)
+    end
+
+    context 'when response is nil' do
+      let(:folio_client) { instance_double(FolioClient, patron_info: nil) }
+
+      it 'notifies honeybadger' do
+        patron = described_class.find('foo')
+        expect(patron.patron_info).to be_blank
+        expect(Honeybadger).to have_received(:notify)
```
dnoneill, line 31: Can you add a test that makes sure that when you call patron.patron_info.user_info it doesn't break everything?

corylown:
@jcoyne do we need to handle a nil `patron_info_response` here as well? https://github.com/sul-dlss/mylibrary/blob/main/app/controllers/application_controller.rb#L19.

I also now suspect that this error https://app.honeybadger.io/projects/62116/faults/108594545 is related because it's the same patron and indicates we're getting some kind of error from `FolioGraphqlClient#patron_info`

spec/models/folio/patron_spec.rb
dnoneill: It doesn't seem to be. I am trying to log in locally and I am getting   [HTTP] < 401 Unauthorized application/json (3.576ms) from patron_info. So the fix isn't working.

dnoneill: It looks like self.find isn't being called and patron_info is being passed in from somewhere.

dnoneill: Actually might be a different issue, because I am having problems with connecting to folio.

jcoyne:
@corylown I agree that there's a deeper issue here. That's why there is a Honeybadger.notify.

corylown:
@jcoyne I don't think the notify you're adding here is going to help with this error because we're not setting up the patron using `Folio::Patron.find`. It's getting created using `Folio::Patron.new` and we're passing in data from `FolioGraphqlClient#patron_info`.

jcoyne:
@corylown do we need something like this first? https://github.com/sul-dlss/mylibrary/pull/1124

corylown (approved):

PR #1123, opened 2024-06-18
Revert Shibboleth Warden::Strategies use env['uid']

corylown:


dnoneill:
Maybe we can't use remote_user?

corylown:
> Maybe we can't use remote_user?

We can. We just need to keep uid for the non-dev warden setup.

PR #1189, opened 2024-12-03
Handle non-successful response from borrowdirect

corylown:


app/services/borrow_direct_reshare_client.rb
```diff
@@ -25,7 +27,10 @@ def get(path, **)
   end
 
   def get_json(path, **)
-    parse(get(path, **))
+    parse(check_response(get(path, **)))
+  rescue BorrowDirectError => e
```
cbeer, line 31: Is the client the right place to suppress this error, or should it be rescued at a higher level? (like, say, https://github.com/sul-dlss/mylibrary/blob/6d7795224927cea06d506a776bcce823bb0998e9/app/models/borrow_direct_reshare_requests.rb#L20?)

I guess I'm worried rescuing the errors this low in the stack could make it harder to debug problems.

corylown: Makes sense. I moved the rescue up to `BorrowDirectReshareRequests#reshare_requests`.

PR #1190, opened 2024-12-04
Upgrade to Rails 7.2

jcoyne:


public/icon.png
dnoneill, line 1: Can we remove these icon files?

dnoneill:
@jcoyne I have tested this locally so it looks to be working so I think the only thing that should get changed before merge is the public/icon.png and public/icon.svg files should be deleted.

PR #1199, opened 2025-01-10
update ci to 3.4

dnoneill:


.github/workflows/ruby.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby: ["3.2", "3.3"]
+        ruby: ["3.3", "3.4"]
```
jcoyne, line 14: Mylibrary is currently deployed on 3.2, so we should keep that until the deploy is on something newer.

PR #1244, opened 2025-04-21
Default to an empty item rather than nil

### LINKED ISSUE #1240: [mylibrary/prod] NoMethodError: undefined method '[]' for nil
## Backtrace

    line 30 of [PROJECT_ROOT]/app/models/concerns/folio/folio_record.rb: Folio::FolioRecord#shelf_key
    line 104 of [PROJECT_ROOT]/app/models/folio/request.rb: Folio::Request#sort_key
    line 13 of [PROJECT_ROOT]/app/controllers/checkouts_controller.rb: block in CheckoutsController#index

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/62116/faults/119389956)

taylor-steve:
Closes #1240?

Likely hitting this item method and not the one in the concern?

### DIFF
--- a/app/models/folio/request.rb
+++ b/app/models/folio/request.rb
@@ -124,7 +124,7 @@ module Folio
     def manage_request_link; end
 
     def item
-      record.dig('item', 'item')
+      record.dig('item', 'item') || {}
     end
   end
 end


corylown (commented):
Should we just remove this method and rely on the one defined in the concern?

corylown:
Going to merge as is to get the fix in then will follow up with some cleanup.

corylown (approved):

taylor-steve:
Yes we should. I did this right before I got on the plane and didn't have time to unpack the private vs public method

PR #1272, opened 2025-07-29
Update field for sending payment account id stubs

jgreben:
@thatbudakguy I spoke with the Cybersource support team. In Mylibrary test or dev (whichever one you were using before that is connected to the Cybersource test instance) can we switch the uuid stub field that we use from `complete_route` to `merchant_defined_data1`? According to [this reporting document (page 88)](https://developer.cybersource.com/content/dam/docs/cybs/en-us/reporting/user/all/ebc/reporting-ug.pdf) it says that MerchantefinedData_field1  is actually 1175 chars. Maybe we can test that out using this PR?


jgreben:
@thatbudakguy let me know what you think of the changes here in order to try to use more than one merchant_defined_data# field. 

It turns out I am not able to deploy to the dev servers, so if you think these changes will work, can you please deploy when you get a chance and I can test it out. Thanks.

PR #1364, opened 2026-05-04
Convert checkouts partial into a component

cbeer:
(not a very good component, but at least a starting point for lifting it over to sul-requests soon)


marlo-longley (approved):
Wow, mylibrary's first ViewComponent!

PR #1376, opened 2026-05-07
Distinguish between user data from mod-users vs graphql

cbeer:
This is a kinda silly seam in mylibrary, but should make lifting this code over to requests a little clearer.

### DIFF
--- a/app/models/folio/patron.rb
+++ b/app/models/folio/patron.rb
@@ -3,27 +3,37 @@
 module Folio
   # Class to model Patron information
   class Patron
-    attr_reader :patron_graphql_response
+    attr_reader :user_info
 
-    def initialize(patron_graphql_response)
+    def initialize(user_info = nil, patron_graphql_response: nil)
+      if user_info.nil? && patron_graphql_response.nil?
+        raise ArgumentError,
+              'Must provide either user_info or patron_graphql_response'
+      end
+
+      @user_info = user_info || patron_graphql_response&.dig('user') || {}
       @patron_graphql_response = patron_graphql_response
+    end
+
+    def patron_graphql_response
+      @patron_graphql_response ||= FolioClient.new.patron_info(user_info['id'])
+    end
+
+    def extended_user_info
+      patron_graphql_response['user']
     end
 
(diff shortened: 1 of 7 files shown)


app/models/folio/patron.rb
```diff
@@ -3,27 +3,37 @@
...
+      @user_info = user_info || patron_graphql_response&.dig('user') || {}
       @patron_graphql_response = patron_graphql_response
     end
 
+    def patron_graphql_response
+      @patron_graphql_response ||= FolioClient.new.patron_info(user_info['id'])
+    end
+
+    def extended_user_info
+      patron_graphql_response['user']
```
taylor-steve, line 23: Looking at the old `user_info`, do we need to `|| {}` or was that guard unnecessary?

taylor-steve (commented):
I'm having a little trouble following so I'm unclear, but does https://github.com/sul-dlss/mylibrary/blob/63ec2abe7186fa0cc064236b1a3e923390cb85bd/app/models/folio/patron.rb#L245 also need to use `extended_user_info` (or otherwise change)? Seems like that key only exists in the graphql response.

marlo-longley:
I was also just reviewing this. I wanted to ask if this is the intention/upshot:

If you never call a method that needs `extended_user_info` (like `barred?`) on a FOLIO/REST-initialized patron, GraphQL is never hit, it saves us interacting with GraphQL. So in Requests if you are just logging in but not getting all the MyLibrary type account data, we may not need to hit GraphQL?

cbeer:
Ugh. `patronGroupId` is our own invention because upstream just calls it `patronGroup` (which we use for the... patron group) 🙄

app/models/folio/patron.rb
cbeer: I think that guard was unnecessary, maybe just for convenience in stubbing for tests?

cbeer:
>  So in Requests if you are just logging in but not getting all the MyLibrary type account data, we may not need to hit GraphQL?

Yes, that's my ultimate goal.

marlo-longley (approved):

